PR #7, opened 2015-02-14
support for user registration and authentication with ORCiD

### LINKED ISSUE #1: Investigate ORCID as a potential identity provider / a&a tool
This is probably a good place to start: http://orcid.org/organizations/integrators

Check if a BIP user may register with ORCiD account and how can we support that process in devise/omniauth. Report findings here.

Nuanda:
Fixes #1 

A lot of stuff here. Some comments:
 - using omniauth with orcid strategy to auth a user
 - sessions_controller set to do the auth (as a OAuth callback)
 - orcid_client service added to use ORCiD Public API to retrieve user's info based on her/his uid
 - parsing that output is the reason I use nokogiri
 - orcid.yml provides all public orcid config (I may switch testing env config to orcid sandbox after they verify my email there - there are some serious emailing problems with their email verification - but it's ok as it is even now)
 - user is persisted on successful orcid callback even if further user's info retrieval was a failure

### DIFF
--- a/config/secrets.yml
+++ b/config/secrets.yml
@@ -13,13 +13,19 @@
 development:
   secret_key_base: 146b411db92206e05af4c02dcb0ac77db0925dff83ca4ecef8434aa8b8ed381203ae3c4cd809d684ec9a9108054b0a1621b9f409c6babf904e0f876a129f56c8
   devise_secret_key: 2cb5fde01129a9598d708352b78d62b5a7a7ac96de19776e1b963b9a6b727684539aae763b59ed84c5ea7e3e62a05a35da97c817cc79e8bcf605d6364b27d220
+  orcid_key: <%= ENV["ORCID_KEY"] %>
+  orcid_secret: <%= ENV["ORCID_SECRET"] %>
 
 test:
   secret_key_base: dc681d4add36415b18f7b6b4437622f306beee43009105c8ddc98ca88729a939d866bbc78923b454749f9999b4df07ad6d71c566395cb38f4c1179ff6965b31b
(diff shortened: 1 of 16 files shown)


config/secrets.yml
```diff
@@ -13,13 +13,19 @@
 development:
   secret_key_base: 146b411db92206e05af4c02dcb0ac77db0925dff83ca4ecef8434aa8b8ed381203ae3c4cd809d684ec9a9108054b0a1621b9f409c6babf904e0f876a129f56c8
   devise_secret_key: 2cb5fde01129a9598d708352b78d62b5a7a7ac96de19776e1b963b9a6b727684539aae763b59ed84c5ea7e3e62a05a35da97c817cc79e8bcf605d6364b27d220
```
nowakowski: Should these secrets be held in the repo?

config/secrets.yml
```diff
@@ -13,13 +13,19 @@
...
+  orcid_key: <%= ENV["ORCID_KEY"] %>
+  orcid_secret: <%= ENV["ORCID_SECRET"] %>
 
 test:
   secret_key_base: dc681d4add36415b18f7b6b4437622f306beee43009105c8ddc98ca88729a939d866bbc78923b454749f9999b4df07ad6d71c566395cb38f4c1179ff6965b31b
   devise_secret_key: b5b4d61cbb62a63cd2d5f3ef7c22811a5d54ba8f51a9ad9e1a790368e73f3d2244d72ed3d0e777cf9cbce3372ffb9694896dbbd5152709979aa5779031ba914e
```
nowakowski: See above.

config/secrets.yml
Nuanda: For dev/test - yes. Just make sure no prod things are in repo.

config/secrets.yml
Nuanda: This is actually exactly the opposite - these are NOT held in repo. That is precisely why this is secure :).

nowakowski:
:+1:

PR #17, opened 2015-02-26
Plant populations data grid

Nuanda:
Replaces the current temporary (PlantVarieties) data with PlantPopulations data in the main data grid. Made to resemble http://www.interstoredb.org/Brassica/cs/ - one thing that is missing is the Maps. The other, of course, the excel export.


Nuanda:
I have rebuilt the data grid strategy so it works by ajax - the PP/PL index actions only render the table header and then run datatables to load grid_data with ajax. For the moment we have client-side processing (faster and data is not that large) but in same cases we will also need server-side processing (too much data for client to handle).

Please review the code.

Nuanda:
Finally, some improvements in the js code for better dT UX.

app/controllers/plant_populations_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      data: data
+    }
+
+    render json: response, layout: false
+  end
+
+
+  private
+
+  def meaningless(value)
```
kammerer: `meaningless?` maybe?

app/models/plant_line.rb
```diff
@@ -19,4 +19,19 @@ class PlantLine < ActiveRecord::Base
                           foreign_key: 'plant_line_name',
                           association_foreign_key: 'plant_population_id'
 
-end
\ No newline at end of file
+  scope :grid_data, ->(ids) do
```
kammerer: Perfect case for a decorator / presenter ;) I won't insist on that but I do not think it should be a scope since it does not return `ActiveRecord::Relation`. It could be a scope with `select` instead of `pluck`.

app/models/plant_line.rb
```diff
@@ -19,4 +19,19 @@ class PlantLine < ActiveRecord::Base
...
+      'common_name',
+      'previous_line_name',
+      'date_entered',
+      'data_owned_by',
+      'organisation'
+
+    where(plant_line_name: ids).
+      joins(:taxonomy_term).
+      order(:plant_line_name).
+      pluck(columns.join(','))
```
kammerer: Or `pluck(*columns)`.

spec/models/plant_line_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe PlantLine, type: :model do
+  describe '#grid_data' do
+    it 'returns empty result when no plant lines found' do
+      expect(PlantLine.grid_data([1])).to be_empty
+    end
+
+    it 'orders populations by common name' do
+      pl1 = create(:plant_line)
```
kammerer: You could use `create_list(:plant_line, 3)`

app/controllers/plant_lines_controller.rb
```diff
@@ -0,0 +1,17 @@
+class PlantLinesController < ApplicationController
+  def index
+  end
+
+  def data_grid
```
kammerer: I would use index action with response depending on request Accept header for more RESTful API.

kammerer:
Not sure why two of my comments seem to be off by one line...

kammerer:
I think that having at least one request spec for each controller would be a good thing.

Nuanda:
Implemented all remarks, apart from the 'decorator' one - more about that one below.

Nuanda:
I agree having a scope that returns just an array is wrong (also, I guess having the group clause in a scope does not make sense, either), and I will fix that one way or the other, though it is not trivial to choose the proper way. We should consider the following:

1. Datatables expect plain array of data. The-rails-way would be to retrieve appropriate range of model objects, then make a full pass and turn them into an array. This is however a waste of time, since all that could be easily done by the DB through a direct pluck query. So, I can extract things like ordering or dummy records drop into chainable scopes, and then add the pluck query on top of them.

2. Is it really proper to put pluck/group arel code inside a decorator? This somehow does not seem ok to me.

I will extract non-scope code into model class operations, for now, while I wait your further opinion.

kammerer:
1. I do not think that the Rails way disallows fetching tabular data from DB when you have a reason to do so.
2. I think it is better do that in decorator than in controller but the latter is still quite common. Seriously though, I do not think this is super important in an app of this size. It could be done in a class method on the model.

Nuanda:
@kammerer please see if that could be merged.

Issue #16, opened 2015-02-25
Gramene Taxonomy for Plant Line genus - species - subtaxa definition


Nuanda:
(related to #12)

I have parsed current GR_tax into the system and tried to check which PlantLines could be defined with it. Here is the matching table for species:

present in GR_tax: ["", "alboglabra", "balearica", "barrelieri", "bivoniana", "carinata", "cretica", "deflexa", "drepanensis", "elongata", "fruticulosa", "gravinae", "hilarionis", "incana", "insularis", "juncea", "macrocarpa", "maurorum", "montana", "napus", "nigra", "oleracea", "oxyrrhina", "perviridis", "procumbens", "rapa", "repanda", "rupestris", "souliei", "spinescens", "tournefortii", "villosa"]

absent in GR_tax: ["unspecified", "not applicable", "none", "atlantica", "bourgaei", "oleracia", "bourgeaui", "desnottesii", "nivalis", "juncea?", "rapa and napus", "rapa?", "nigra?", "napus?", "hybrid", "MIXED"]

Issues:
 - "bourgeaui", "oleracia" - simple typos?
 - "rapa and napus", "hybrid", "MIXED" - leave as no species defined?
 - "juncea?", "rapa?", "nigra?", "napus?" - GR_tax do not accommodate for questionable classification: should we assign it anyway but model questionable assignment with a boolean field?
 - "atlantica", "bourgaei", "desnottesii", "nivalis" - four species that are simply not present in GR_tax - file  an addition to Gramene?

Anyway, species is relatively easy :). Going to subtaxa we have bigger problems.

present in GR_tax: ["", "abyssinica", "acephala", "alboglabra", "amplexicaulis", "bivoniana", "botrytis", "capitata", "carinata", "chinensis", "costata", "cretica", "drepanensis", "elongata", "fruticulosa", "gemmifera", "gongylodes", "incana", "insularis", "italica", "juncea", "macrocarpa", "napobrassica", "napus", "narinosa", "nigra", "nipposinica", "oleifera", "oleracea", "parachinensis", "pekinensis", "perviridis", "rapa", "spinescens", "villosa"]

[...]

PR #25, opened 2015-02-28
Better pl tt curation

Nuanda:
Added a lot of non-canonical terms from CS to the TT table. Simplified taxonomy curation task. Now we relatively carefully match all PLs except 103 cases - see #16 issue for the current state.

### DIFF
--- a/lib/tasks/curation_tt.rake
+++ b/lib/tasks/curation_tt.rake
@@ -3,117 +3,44 @@ namespace :curate do
   task plant_taxonomy: :environment do
     # Now comes the hard part: we need to heuristically link existing PL records
     # to TTs based on the content of their species, genus and subtaxa fields
-    pls = PlantLine.all;
-    @cntr = 0
-    @success_cntr = 0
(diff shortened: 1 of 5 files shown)


lib/tasks/curation_tt.rake
```diff
@@ -3,117 +3,44 @@ namespace :curate do
...
+
+      cntr += 1
+      tt = TaxonomyTerm.find_by(name: name_to_find)
+      if tt
+        pl.taxonomy_term = tt
         pl.save
-        @success_cntr += 1
-        @cntr += 1
-        next pl
+        success_cntr += 1
```
nowakowski: This task takes a long time to execute. Consider adding some on-the-fly progress messages to indicate how things are going (at least show how many records have been processed, e.g. `if @cntr % 1000 == 0 puts "Processed #{cntr.to_s} records..." end`).

Nuanda: Added. Please note you do not need to explicitly call to_s when interpolating.

nowakowski:
:+1:

PR #33, opened 2015-03-11
Better plant lines submission names

### LINKED ISSUE #29: Make submissions more population-oriented.
Currently we don't know how the submission module will be decomposed (see #24), but for the first prototype I propose to assume, for the time being, that user can only submit plant populations.

Therefore:
- [x] change button label "Submit new research data" to "Submit plant population"
- [ ] change the information in the submissions list items to what I describe below
- [x] change each step title from "Submission details" to "Plant population submission"

The proposition of the submission item list information:

Population: BraCKDH (B. rapa). DH segregating. Parents: Chiifu type 401-42 | Kenshin type 402-43.

That is population name, species, type (if present), parents (if present). Then "submitted on (date)" or "last modification (date)", depending on the current submission state and the "submitted" label or actions buttons, like it is shown now.

kammerer:
* show 'B.' instead of 'Brassica'
* do not show empty "Parents: " string if not parent lines are selected

Fixes #29

### DIFF
--- a/app/decorators/plant_population_submission_decorator.rb
+++ b/app/decorators/plant_population_submission_decorator.rb
@@ -7,7 +7,7 @@ class PlantPopulationSubmissionDecorator < Draper::Decorator
       l << " (#{species_name})" if species_name
       l << "." if population_name || species
       l << " #{population_type}." if population_type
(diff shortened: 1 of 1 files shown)


Nuanda:
:+1:

PR #35, opened 2015-03-13
CropStore schema migration fix

nowakowski:
The autogenerated CropStore schema migration had several problems which prevented certain ActiveRecord joins from functioning correctly. The issue is associated with treatment of primary keys in RoR migrations (by default RoR assumes - incorrectly in our case - that the primary key of a table is of type INT; for all tables without an explicitly defined primary key an `id` column is automatically generated).


Nuanda:
:+1:

Issue #37, opened 2015-03-13
Wrong uniqueness index in the PPL joining table


Nuanda:
The current migration says:

     create_table "plant_population_lists", primary_key: "plant_population_id", force: :cascade do |t|
        t.text "plant_line_name",   default: "unspecified", null: false

I believe this is wrong, it prevents assigning two different PLs to the same PP.

Nuanda:
I think it's even worse :(. There are TWO migrations creating that table - both probably equally wrong.

nowakowski:
Yes, this duplication is strange. Fixed.

PR #38, opened 2015-03-13
New pp data grid

Nuanda:
This introduces:
 - different, more general handling of PL.grid_data queries
 - uses strong parameters to prohibit querying too freely
 - changes in PP data grid display


Nuanda:
@kammerer This is not ready for merging, since I wait for #37 fix, but please take a look, esp. at PlantLine.grid_data, PlantLinesController and data_grid.js.coffee to check the proposed changes in the data grid querying mode. We might try to replicate this way of filtering to other models as well, even maybe abstract away the :query switch handling (when such a need arises).

I left your :search handling for autocomplete as it was.

Tell me if this is a good way to go.

db/migrate/20150222094809_create_taxonomy_terms.rb
```diff
@@ -2,7 +2,7 @@ class CreateTaxonomyTerms < ActiveRecord::Migration
   def change
     create_table :taxonomy_terms do |t|
       t.string :label, null: false, index: true
-      t.string :name, null: false
+      t.string :name, null: false, index: true
```
kammerer: Don't we want to make it a unique index?

Nuanda: Good point, this should de facto be a unique value. Otherwise there really is no point in having a taxonomy. Will fix it in another issue/PR.

lib/tasks/obo.rake
```diff
@@ -19,5 +20,116 @@ namespace :obo do
...
+    TaxonomyTerm.
+      where(name: variety_names(genus, species, subtaxa)).
+      first
+  end
+
+  def save_with_parent(term, parent)
+    term.canonical = false
+    term.label = 'CROPSTORE'
+    term.parent = parent
+    term.save
```
kammerer: Minor thingie, most likely not important in this case - I like to use bang version by default when saving records (`save!`, `create!`, `update_attributes!`) unless I am checking return value or validity.
Do not mind me if I mentioned that already, I do that often ;)

Nuanda: Fixed in latest commit.

db/migrate/20150222094809_create_taxonomy_terms.rb
Nuanda: Addressed in #45

Nuanda:
Ah, ok, I somehow commited _wrong_ schema, the one without Piotr's recent PPL fix. Travis relies on schema for testing.

app/models/plant_line.rb
```diff
@@ -30,10 +30,14 @@ def self.grid_data(filter)
       'data_owned_by',
       'organisation'
 
-    query = where(plant_line_name: filter[:plant_line_names]) if filter[:plant_line_names].present?
+    query = where(filter[:query]) if filter[:query].present?
     query = where('plant_line_name ILIKE ?', "%#{filter[:search]}%") if filter[:search].present?
     query ||= none
 
+    filter[:query].each do |k,_|
```
kammerer: Did you introduce it because you will be adding querying by many other fields? I do not see anything wrong with it provided you make sure `filter[:query]` is limited to a `Hash`, otherwise it allows SQL injection (I know it is checked in the controller, I wonder if it should not be checked here too).

kammerer:
:+1:

app/models/plant_line.rb
Nuanda: Well, we indeed might need multiple-criteria testing (I made a spec for that) but, what is more important, I plan to refactor this to make that code work cross-model - for every model class that should be datatablesable ;). Having it generic helps.

For the time being I rely on strong parameters here - and I also added a spec for that (testing on PL case) - but I admit this might make it vulnerable for e.g. future API calls (if not outfitted with strong parameters). Let me consider  additional check after I abstract this code up, since the check will be model-specific. Setting another issue for that.

PR #45, opened 2015-03-14
making sure no taxonomy term gets duplicated

### LINKED ISSUE #44: Make TT name unique with index and validation, add tests

Nuanda:
Fixes #44


kammerer:
Hmm, you commited a whole lot of changes in db/schema.rb...

Nuanda:
I know, but that's not me, really. @nowakowski way of dealing with legacy data means we have quite different schemas depending whether we run migrations on TEST or on DEV - both update schema.rb. Sorry for that.

PR #49, opened 2015-03-14
using strong params to prevent SQL injection in model

### LINKED ISSUE #47: Make grid_data model call invulnerable for sql injection by permitted parameters filtering

Nuanda:
Fixes #47

### DIFF
--- a/app/models/plant_line.rb
+++ b/app/models/plant_line.rb
@@ -30,17 +30,28 @@ class PlantLine < ActiveRecord::Base
       'data_owned_by',
       'organisation'
 
-    query = where(filter[:query]) if filter[:query].present?
+    safe_query = grid_data_params(filter[:query])
+    query = where(safe_query) if safe_query.present?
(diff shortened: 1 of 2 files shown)


Nuanda:
@kammerer Please, have a look. This is how I propose to move the strong params guard from controller layer to model layer, for that particular grid_data call. Left all strong params in PLController for now.

Your action is needed here:
 - decide whether to leave :search autocomplete functionality in PL.grid_data method
 - if yes, move entire grid_data_params code from PLController to PL model
 - if no, write separate model scope/method for autocomplete and shorten PLController.grid_data_params by :query parameters, leaving :search parameter
 - write an example or two for that autocomplete functionality in plant_line_spec, for safer refactoring in the future.

kammerer:
I think it makes sense and can be merged to master. I am not sure about :search yet, I have an idea I want to test but I can do that in a new branch.

Nuanda:
All right - I'll file you a ticket so we don't forget to address it. Currently strong parameters are done twice, what is unnecessary, esp. for autocomplete. For now - merging this.

PR #54, opened 2015-03-15
ActiveRecord classes for the "yellow", "red" and "orange" portions of the CropstoreDB data model.

nowakowski:
Placeholder PR. Do not merge yet.

### DIFF
--- a/app/models/genotype_matrix.rb
+++ b/app/models/genotype_matrix.rb
@@ -1,0 +1,5 @@
+class GenotypeMatrix < ActiveRecord::Base
+
+  belongs_to :linkage_map, foreign_key: 'linkage_map_id'
+
+end
--- a/app/models/genotype_matrix.rb
+++ b/app/models/genotype_matrix.rb
@@ -1,5 +1,5 @@
 class GenotypeMatrix < ActiveRecord::Base
 
-  belongs_to :linkage_map, foreign_key: 'linkage_map_id'
+  belongs_to :linkage_map
 
 end
--- a/app/models/linkage_group.rb
+++ b/app/models/linkage_group.rb
@@ -1,12 +1,11 @@
 class LinkageGroup < ActiveRecord::Base
 
   has_and_belongs_to_many :linkage_maps,
-                          join_table: 'map_linkage_group_lists',
-                          foreign_key: 'linkage_group_id'
+                          join_table: 'map_linkage_group_lists'
 
   has_many :map_positions
 
-  has_many :map_locus_hits, foreign_key: 'linkage_group_id'
+  has_many :map_locus_hits
 
   has_many :qtls
 
--- a/app/models/linkage_group.rb
+++ b/app/models/linkage_group.rb
@@ -9,4 +9,6 @@ class LinkageGroup < ActiveRecord::Base
 
   has_many :map_locus_hits, foreign_key: 'linkage_group_id'
 
+  has_many :qtls
+
 end
--- a/app/models/linkage_group.rb
+++ b/app/models/linkage_group.rb
@@ -1,0 +1,12 @@
+class LinkageGroup < ActiveRecord::Base
+
+  has_and_belongs_to_many :linkage_maps,
+                          join_table: 'map_linkage_group_lists',
+                          foreign_key: 'linkage_group_id',
+                          association_foreign_key: 'linkage_map_id'
+
+  has_many :map_positions
+
+  has_many :map_locus_hits, foreign_key: 'linkage_group_id'
+
+end
--- a/app/models/linkage_group.rb
+++ b/app/models/linkage_group.rb
@@ -2,8 +2,7 @@ class LinkageGroup < ActiveRecord::Base
 
   has_and_belongs_to_many :linkage_maps,
                           join_table: 'map_linkage_group_lists',
-                          foreign_key: 'linkage_group_id',
-                          association_foreign_key: 'linkage_map_id'
+                          foreign_key: 'linkage_group_id'
 
   has_many :map_positions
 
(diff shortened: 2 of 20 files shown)


nowakowski:
I'll split this into two parts to avoid overlarge PRs. @Nuanda could you review?

app/models/genotype_matrix.rb
```diff
@@ -0,0 +1,5 @@
+class GenotypeMatrix < ActiveRecord::Base
+
+  belongs_to :linkage_map, foreign_key: 'linkage_map_id'
```
Nuanda: Is it required to specify FKey in this form? Isn't it so by default?

app/models/linkage_group.rb
```diff
@@ -0,0 +1,14 @@
+class LinkageGroup < ActiveRecord::Base
+
+  has_and_belongs_to_many :linkage_maps,
+                          join_table: 'map_linkage_group_lists',
+                          foreign_key: 'linkage_group_id',
+                          association_foreign_key: 'linkage_map_id'
```
Nuanda: I assume you do not want to rename these keys to traditional SQL values? For me it's ok since I hope to work with relations only, but could be difficult in the future when debugging problems directly on the DB...

app/models/plant_population.rb
```diff
@@ -10,6 +10,12 @@ class PlantPopulation < ActiveRecord::Base
 
   has_many :plant_population_lists, foreign_key: 'plant_population_id'
 
+  has_many :linkage_maps, foreign_key: 'mapping_population'
+
+  has_many :population_loci, foreign_key: 'plant_population'
+
+  has_many :processed_trait_datasets, foreign_key: 'population_id'
+
```
Nuanda: I know this is a legacy system but you'll admit this looks terrible... three FKeys to the same table with three different names... :(

app/models/qtl_job.rb
```diff
@@ -0,0 +1,6 @@
+class QtlJob < ActiveRecord::Base
+  self.table_name = 'qtl_jobs'
```
Nuanda: Is that really needed? Looks rather regular to me.

Nuanda:
A can't check right now whether this works or not, but apart from these simple comments regarding keys, I think it is ok.

app/models/genotype_matrix.rb
nowakowski: Fixed.

app/models/linkage_group.rb
nowakowski: Don't want to cram too many changes into one PR - for now I'm setting up model classes; you are right in that some FK names should be fixed but it's not as simple as renaming columns (I'll need to generate additional numerical ID columns and set them up as keys since the existing keys are strings). I intend to do this as a separate task. You are correct about the association FK designation; that's been fixed.

app/models/plant_population.rb
nowakowski: It looks very ugly but see above. :)

app/models/qtl_job.rb
nowakowski: Yeah, originally I wanted to name the class QTLJob - hence the "nonstandard" table name - then I thought better of it so now it isn't really needed. I'll remove it.

nowakowski:
Fixed, rebased to latest master and committed. @Nuanda - if you have no further comments I think we can merge this PR and close the branch.

app/models/linkage_group.rb
```diff
@@ -0,0 +1,13 @@
+class LinkageGroup < ActiveRecord::Base
+
+  has_and_belongs_to_many :linkage_maps,
+                          join_table: 'map_linkage_group_lists',
+                          foreign_key: 'linkage_group_id'
+
+  has_many :map_positions
+
+  has_many :map_locus_hits, foreign_key: 'linkage_group_id'
```
Nuanda: If I am right, you don't need to specify foreign_key: 'linkage_group_id' here or in the habtm above, either.

app/models/map_locus_hit.rb
```diff
@@ -0,0 +1,8 @@
+class MapLocusHit < ActiveRecord::Base
+
+  belongs_to :linkage_map, foreign_key: 'linkage_map_id'
+  belongs_to :linkage_group, foreign_key: 'linkage_group_id'
```
Nuanda: The two fkey declarations above are not required, right?

app/models/linkage_map.rb
```diff
@@ -0,0 +1,14 @@
+class LinkageMap < ActiveRecord::Base
+
+  has_and_belongs_to_many :linkage_groups,
+                          join_table: 'map_linkage_group_lists',
+                          foreign_key: 'linkage_map_id',
+                          association_foreign_key: 'linkage_group_id'
```
Nuanda: I'd guess both these fkey and assoc_fkey declarations are not required.

app/models/linkage_group.rb
```diff
@@ -0,0 +1,13 @@
+class LinkageGroup < ActiveRecord::Base
+
+  has_and_belongs_to_many :linkage_maps,
+                          join_table: 'map_linkage_group_lists',
+                          foreign_key: 'linkage_group_id'
```
Nuanda: Explicit fkey declaration not required?

app/models/linkage_group.rb
Nuanda: Explicit fkey declaration not required?

Nuanda:
Just spotted some, possibly surplus, fkey declarations. If I'm right and they are not required, please remove them. Afterwards it's :+1: and you are free to merge.

PR #60, opened 2015-03-19
Plant line list in submission

kammerer:
This PR introduces the following:

1. Allows to pick existing PLs in 3rd step
2. Allows to define new PLs in 3rd step
3. Stores submitted data as 'regular' objects (PP, PLs, PPLs)

Surely there is a lot of places which need work but I think it is better to address these in separate issues.


app/assets/javascripts/jquery.scrollTo.js
```diff
@@ -0,0 +1,187 @@
+/*!
```
Nuanda: Shouldn't we put external libraries to vendor/assets/javascripts ?

app/assets/javascripts/submissions.coffee
```diff
@@ -16,50 +16,76 @@ plantLineSelectOptions = ->
...
+    $input = $("<input type='hidden' name='submission[content][new_plant_lines][][" + attr + "]' />")
+    $input.val(val)
+    $input.appendTo($container)
+  )
 
-switchTarget = (button) ->
-  markActiveButton(button)
-  activateTargetFields(button)
+newPlantLineForListContainerId = (plant_line_name) ->
+  'new-plant-line-' + plant_line_name.split(/\s+/).join('-').toLowerCase()
```
Nuanda: Are that complex processing required due to not having a proper integer id for a PL? This is why we need to rely on plant_line_name for identification?

app/models/plant_line.rb
```diff
@@ -45,6 +45,9 @@ def self.grid_data(filter)
       .pluck(*columns)
   end
 
+  def self.genetic_statuses
+    order('genetic_status').pluck('DISTINCT genetic_status').reject(&:blank?).reject { |s| ['none', 'not applicable', 'unspecified'].include?(s) }
```
Nuanda: Well, I'm not a 80-char zealot ;) but this is just a bit too much. Try to wrap it.

app/views/submissions/steps/_step03.html.haml
```diff
@@ -17,4 +17,64 @@
...
+        %p.help-block Optional
+
+      .action-buttons
+        %button.cancel-new-plant-line-for-list{type: 'button'} Cancel
+        %button.add-new-plant-line-for-list{type: 'button'} Confirm
+
   = render partial: 'submissions/steps/buttons', locals: { f: f, submission: submission }
+
+  - @content.new_plant_lines.each do |plant_line|
+    %div{id: "new-plant-line-#{plant_line.plant_line_name.split.join('-').downcase}"}
```
Nuanda: Again the "plant_line_name as a unique id" problem?

app/views/submissions/steps/_step03.html.haml
```diff
@@ -17,4 +17,64 @@
...
+      = hidden_field_tag "submission[content][new_plant_lines][][plant_line_name]", plant_line.plant_line_name
+      = hidden_field_tag "submission[content][new_plant_lines][][taxonomy_term]", plant_line.taxonomy_term
+      = hidden_field_tag "submission[content][new_plant_lines][][common_name]", plant_line.common_name
+      = hidden_field_tag "submission[content][new_plant_lines][][previous_line_name]", plant_line.previous_line_name
+      = hidden_field_tag "submission[content][new_plant_lines][][comments]", plant_line.comments
+      = hidden_field_tag "submission[content][new_plant_lines][][genetic_status]", plant_line.genetic_status
```
Nuanda: Have you considered putting that in a array, with reflection 'send' call on plant_line for each element?
Like:
%w(plant_line_name taxonomy_term ...).each do |attr|
 = hidden_field_tag "submission[content][new_plant_lines][][#{attr}]", plant_line.send(attr)

app/services/submission/plant_population_finalizer.rb
```diff
@@ -0,0 +1,71 @@
...
+    %i[female_parent_line male_parent_line].each do |parent_line_attr|
+      if parent_line = PlantLine.find_by(plant_line_name: submission.content.step03[parent_line_attr])
+        attrs.merge!(parent_line_attr => parent_line)
+      end
+    end
+
+    attrs.merge!(submission.content.step01.to_h)
+    attrs.merge!(submission.content.step02.to_h)
+    attrs.merge!(submission.content.step04.to_h)
+    attrs.delete(:name) # FIXME change :name to :plant_population_name in the form
```
Nuanda: I don't understand that thing. What really should be changed in the model and how?

Nuanda:
Made some minor remarks.

app/assets/javascripts/jquery.scrollTo.js
kammerer: Makes sense

app/assets/javascripts/submissions.coffee
kammerer: No, this code handles PLs added during the submission, they are not stored as regular PL records.

app/models/plant_line.rb
kammerer: Ok. BTW 'none', 'not applicable', etc are already a thing of the past?

app/services/submission/plant_population_finalizer.rb
kammerer: Not in the model, in the form class. I used `name` property but using `plant_population_name` would make things slightly simpler.
On the other hand changing `PlantPopulation#plant_population_name` to `#name` in the model would make sense to.

app/models/plant_line.rb
Nuanda: They should be.

tg

Issue #59, opened 2015-03-17
Address the problem of search and query separation for PLs


Nuanda:
See #49 for some additional information.

PR #64, opened 2015-03-21
Do not make references to grid data in model layer

kammerer:
Regarding #59 - I am OK with handling both `:query` and `:search` by the same method (at least at this point). However I am not entirely happy with the name `.grid_data`. In this PR I suggest one possible way to rename and somewhat generalize these methods (not included in the PR is actual usage but now they allow to fetch less data for autocomplete fields).

What do you think?


Nuanda:
It's all right. :+1:

PR #66, opened 2015-03-21
Add missing PP records and link PSU to PP

nowakowski:
Three missing records added ('whole plant', 'average of blocks' and 'average of trays').

Run `rake curate:add_missing_plant_parts` to fix production DB.


nowakowski:
@wjurkowski - could you take a look at the proposed values in `plant_parts` fields and possibly suggest better alternatives? (I'm referring in particular to `description` and all the e-mail addresses).

Nuanda:
I'm not sure what Wiktor thinks about that, but the emails should be either NULL or to someone from TGAC, in my opinion.

nowakowski:
@Nuanda Yes, I hesitated there for a second but decided to tell the truth. I'll happily change it to whatever @wjurkowski suggests.

wjurkowski:
Not sure if I understand this one. Why do we need to propose new values? An where I can see the proposed values? The email address should point to either someone who upload the data, in table that describes the actual experiment/trial. Or the second option is the contact to authority who assigned specific plant systematisation not yet present in ontologies

nowakowski:
@wjurkowski Take a look at the changelog for this branch and in particular at the contents of the `lib/tasks/add_missing_plant_parts.rake` file (it's a Rake task but the code should be fairly self-explanatory). These are the three records which we need to add to `plant_parts` in order to ensure that `plant_scoring_units` has a meaningful foreign key linking to this table. My question related to whether the proposed values for all `plant_parts` columns are okay or whether they should be changed (and if so - how).

Since the only person whose e-mail address is listed for `plant_scoring_units` is **pierre.carion@bbsrc.ac.uk**, I can copy this address over to `plant_parts`; however other fields (such as `description`, `comments`, `data_provenance` etc.) remain a problem.

wjurkowski:
Now I got it! The plant_part field doesn't include all the values already
present in scored_plant_part tables. This is obvious gap.
Of course please add all three fields as you specified.
At the moment we don't have any other suggestions to how the plant parts
could be scored differently. If we will, then it should be added in both.

2015-03-27 11:43 GMT+00:00 Piotr Nowakowski <notifications@github.com>:

nowakowski:
@wjurkowski Affirmative. Merging and closing.

Nuanda:
Again, you shouldn't put our email there - since we have just rewritten these values. They were supplied by someone else - possibly the original author.

Nuanda:
Perhaps you can extract the email value from the records of the table where you took the values from?

wjurkowski:
yes, please put pierre email back. This is also not ideal - as he is not
brassica authority - but he physically inserted the data.

2015-03-27 12:30 GMT+00:00 Tomasz Gubała <notifications@github.com>:

nowakowski:
OK, will do.

Nuanda:
thx

Nuanda:
:+1:

Nuanda:
@nowakowski shouldn't you update that: https://github.com/eSpectrum-IT/brassica/wiki ?

nowakowski:
@Nuanda Done.

Issue #62, opened 2015-03-20
Previous line name is not a reference to plant lines


Nuanda:
@nowakowski Piotr, please confirm that, but the PL's previous_line_name is, actually, not a self-reference to a PL but rather a free text label (of how it was previously called?).

Nuanda:
Ok, I checked it myself since this is kind of urgent.

We have 5781 unique previous_line_names in PLs. These, however, correspond to existing PL plant_line_names only in 25 cases: ["AGDH_1010", "AGDH_3015", "AGDH_3081", "AGDH_3200", "105", "CA91087", "AC2450", "BI88908", "CA91061", "CA97004", "CA91079", "AGDH_3016", "AGDH_4035", "AGDH_4252", "AGDH_5119", "AGDH_5121", "AGDH_6098", "CA91071", "CGN14111", "CA91064", "CA97089", "ECD15/48b", "GK99123", "R-o-18", "102"].

So, clearly it's not a real (self-referencing) FKey.

Nuanda:
Tomek - the current possibility to choose another PL for previous_line_name, when creating a PL, is good, but we also need to allow users to simply put arbitrary text there.

PR #69, opened 2015-03-24
Genetic status input / select alternative

### LINKED ISSUE #63: Better widget for setting genetic status when adding new plant population
User must be able to select one of existing genetic statuses or enter a new one. Currently it is achieved by using a "tagging" field with a limit of 1 entry (implemented with select2 javascript library). This solution has an issue resulting in the field being pre-populated instead of being empty and showing a placeholder with explanation to the user.

kammerer:
This PR intends to fix #63 by introducing separate but mutually exclusive form elements for selecting existing value and entering new value for genetic status. Using any one disables the other.

If accepted the same solution could be used for #62.

### DIFF
--- a/app/assets/javascripts/errors.coffee
+++ b/app/assets/javascripts/errors.coffee
@@ -20,11 +20,11 @@ window.Errors =
     data = {}
     errors = []
 
-    attrs = $(form).find('[name]').not('[type=hidden]').map(-> this.name)
+    attrs = $(form).find('[name]:enabled').not('[type=hidden]').map(-> this.name)
     required = $(form).find('.required, [required]').map(-> this.name)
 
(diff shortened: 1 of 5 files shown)


Nuanda:
Two comments:
 - if one types in his/her own genetic status label, that coincides to be identical to one already in the DB, we probably should (during the final submission) reset it to reference anyway
 - this has probably something to do with the way the "current" submission step state is kept/managed, but when I press Enter somewhere in the Step 3 form (for instance, in the new PL plant line name text field), I go back to Step 2 form.

Anyway, I think this solution is quite all right. :+1:

kammerer:
Genetic status does not seem to be a reference to me.

kammerer:
Fixed issue with form submission being triggered by pressing Enter.

PR #70, opened 2015-03-25
Allow to enter arbitrary string as previous line name in step 3

### LINKED ISSUE #62: Previous line name is not a reference to plant lines

kammerer:
Fixes #62 using the same solution as already used for genetic status.


Nuanda:
:+1:

PR #71, opened 2015-03-26
Front page

Nuanda:
Changes to better compartmentalize code and to structure the front page more or less the way it is designed.

Making it ready for layouting and styling.


kammerer:
:+1:

PR #74, opened 2015-03-27
Navbar and other navigation fixes

Nuanda:


Nuanda:
@kammerer Tomek, please have a look on the application_helper code for navigation bar. I admit this is not a very nice piece of code.

Do you think we should have sth like controller method, returning us the active vav item? Seems more OOP but litter controllers code with nav links names, so I decided to keep all that "knowledge" inside a single helper.

If you have a nicer solution for that, please tell.

Nuanda:
@kammerer Tomek, please have a look on the application_helper code for navigation bar. I admit this is not a very nice piece of code.

Do you think we should have sth like controller method, returning us the active vav item? Seems more OOP but litter controllers code with nav links names, so I decided to keep all that "knowledge" inside a single helper.

If you have a nicer solution for that, please tell.

kammerer:
I think that config/locales is the right place to keep UI strings even if the app is not really internationalized. As for helper / controller choice - I think both are ok.

Nuanda:
Did you mean sth like that?

kammerer:
Yup, :+1:

2015-03-27 17:34 GMT+01:00 Tomasz Gubała <notifications@github.com>:

> Did you mean sth like that?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/74#issuecomment-87000066>.
>



-- 
Tomasz Szymczyszyn

PR #78, opened 2015-03-27
Submission type support

Nuanda:
We need it to differentiate various types of submissions (currently planning to support four).

### DIFF
--- a/db/migrate/20150221153705_create_submissions.rb
+++ b/db/migrate/20150221153705_create_submissions.rb
@@ -1,11 +1,13 @@
 class CreateSubmissions < ActiveRecord::Migration
   def change
     create_table :submissions do |t|
-      t.timestamps null: false
       t.references :user, null: false
       t.string :step, null: false
       t.json :content, null: false, default: {}
-      t.boolean :finalized, null: false, default: false
+      t.boolean :finalized, null: false, default: false, index: true
+      t.integer :submission_type, null: false, index: true
+
+      t.timestamps null: false
     end
 
     add_index :submissions, :user_id
--- a/spec/factories/submission.rb
+++ b/spec/factories/submission.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :submission do
-    submission_type 0
+    submission_type :population
     user
   end
 end
--- a/spec/factories/submission.rb
+++ b/spec/factories/submission.rb
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :submission do
+    submission_type 0
     user
   end
 end
--- a/spec/models/submission_spec.rb
+++ b/spec/models/submission_spec.rb
@@ -6,14 +6,12 @@ RSpec.describe Submission do
   it { should validate_presence_of(:submission_type) }
 
   describe '#submission_type' do
-    let(:submission) { create(:submission) }
+    let(:submission) { build(:submission) }
 
     it 'allows only certain submission type values' do
       %w(population traits qtl linkage_map).each do |t|
         submission.submission_type = t
-        expect(submission.save).to be_truthy
-        submission.reload
-        expect(submission.submission_type).to eq t
+        expect(submission.valid?).to be_truthy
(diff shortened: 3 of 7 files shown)


spec/models/submission_spec.rb
```diff
@@ -2,6 +2,40 @@
...
+      expect { submission.submission_type = 'wrong_submission_type' }.
+        to raise_error ArgumentError
+    end
+
+    it 'honors symbols as type values' do
+      submission.submission_type = :population
+      expect(submission.save).to be_truthy
+    end
+
+    it 'provides handy scopes to query certain types' do
```
kammerer: If these scopes are provided by AR then isn't it like testing AR itself?

spec/factories/submission.rb
```diff
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :submission do
+    submission_type 0
```
kammerer: Why `0` and not `:population`?

spec/models/submission_spec.rb
```diff
@@ -2,6 +2,40 @@
...
+  it { should validate_presence_of(:user) }
+  it { should validate_presence_of(:submission_type) }
+
+  describe '#submission_type' do
+    let(:submission) { create(:submission) }
+
+    it 'allows only certain submission type values' do
+      %w(population traits qtl linkage_map).each do |t|
+        submission.submission_type = t
+        expect(submission.save).to be_truthy
```
kammerer: You may use `build` instead of `create` and `valid?` instead of `save` if there is no real reason to save. Probably no difference right now but when the suite grows it may save some time.

db/migrate/20150221153705_create_submissions.rb
```diff
@@ -1,11 +1,13 @@
 class CreateSubmissions < ActiveRecord::Migration
   def change
     create_table :submissions do |t|
-      t.timestamps null: false
       t.references :user, null: false
       t.string :step, null: false
       t.json :content, null: false, default: {}
-      t.boolean :finalized, null: false, default: false
+      t.boolean :finalized, null: false, default: false, index: true
```
kammerer: Isn't easier to add new migration than modify existing one?

kammerer:
Added some minor comments but in general it is ok. :+1:

db/migrate/20150221153705_create_submissions.rb
Nuanda: Not for me, I mean, with time migrations grow and grow - so I really prefer to have them compact as long as we can - which means up to live beta.

spec/models/submission_spec.rb
Nuanda: I don't like build - it cheats :). I know that in this case it could save a DB hit. I will change it.

spec/factories/submission.rb
Nuanda: This is what a tutorial on booleans in migrations told me - I thought it is not possible to use model-scoped literals here. Will try.

spec/models/submission_spec.rb
Nuanda: Some say it's superfluous but I believe otherwise. Because with time you want to upgrade AR as well.

But in this specific case I also check that no one modifies the four submission enum to sth else. I prefer it stays that way.

db/migrate/20150221153705_create_submissions.rb
kammerer: I do not see any problem with growing number of migrations (I see problem with old migrations which fail but that is quite different thing). However changing existing migration forces other devs to partially recreate their development db... Just my thoughts, I am not saying you should do anything in this case.

spec/models/submission_spec.rb
kammerer: OK, good points.

spec/models/submission_spec.rb
kammerer: I do not think it cheats because I believe that AR object should behave nicely even if they are not saved yet ;)

Issue #51, opened 2015-03-15
Dropping data_status from entire database


Nuanda:
Column names 'data_status' is present in 44 tables. Yet, it always has either of those five values:
 - "published"
 - "public"
 - ""
 - "public_default"
 - "public available under MTA"

Can we assume all that means "public" and remove that column from the DB entirely?

Nuanda:
@wjurkowski Wiktor, do we have your "go ahead" about that?

wjurkowski:
Yes, we assume all data is public. But if its published that's another
matter. Do we have a field somewhere that could store the Pubmed ID of the
paper (if any)?

2015-03-27 10:42 GMT+00:00 Tomasz Gubała <notifications@github.com>:

Nuanda:
About pubmed id i don't know. There was a dedicated cs_publications table in the CS schema, but it was empty. We may, however, introduce a pubmedid field where you find it useful - and the users may supply that value when submitting data. Out of your proposed submissions (populations, traits, qtl, linkage_map) which one could be accompanied by a pubmedid?

Nuanda:
@nowakowski Piotr, please drop data_status field from all the tables it is present in. That means - "a tremendous amount" of tables :).

wjurkowski:
Lets leave it for discussion...

2015-03-27 12:17 GMT+00:00 Tomasz Gubała <notifications@github.com>:

Nuanda:
Sure we can leave the pubmedid discussion open. I asked Piotr to drop the useless (and very frequent!) data_status table.

nowakowski:
Addressed in PR #80.

sarahcayling:
We discussed the pubmed id issue, and there could be pubmed ids on plant trials (related to traits), QTL or on linkage maps. Probably not a population alone without these other things.

Nuanda:
Thanks - I made a note about that in #77

Nuanda:
@nowakowski I still see `data_status` fields in `occasions` and `linkage_groups` tables. Please remove them as well (no useful data inside).

Nuanda:
Related #132

PR #80, opened 2015-03-27
Drop 'data_status' field in model tables

nowakowski:
This PR drops the useless 'data_status' column in all model tables and additionally fixes the bootstrap migration so that it correctly sets up the 'probes' table.


nowakowski:
Fixes #51

Nuanda:
:+1:

Issue #79, opened 2015-03-27
Can we drop 'probes.pcr_yes_or_no'?


nowakowski:
All CropstoreDB records show 'N' in this field, and we have already dropped `pcr_primer_name_a` and `pcr_primer_name_b` due to lack of useful data.

wjurkowski:
yes

2015-03-27 13:44 GMT+00:00 Piotr Nowakowski <notifications@github.com>:

nowakowski:
Fixed in #82

PR #82, opened 2015-03-27
Dropped 'pcr_yes_or_no' from probes

nowakowski:
This field appears useless since its value was 'N' for all records and moreover `pcr_primer_name_a` and `pcr_primer_name_b` had already been dropped.


nowakowski:
Fixes #79

Nuanda:
:+1:

Nuanda:
Fixes #X needs to be in the main PR description when you create it, I think, in order to have it close the relevant issue automatically on merge.

nowakowski:
Didn't know. Thanks.

PR #84, opened 2015-03-27
New submission item

Nuanda:
It looks complex but I tried to add detailed classes so Bartosz is able to style it afterwards. New polymorphic structure should allow us to add new types of submission with less effort.

### DIFF
--- a/spec/decorators/plant_population_submission_decorator_spec.rb
+++ b/spec/decorators/plant_population_submission_decorator_spec.rb
@@ -20,11 +20,14 @@ RSpec.describe PlantPopulationSubmissionDecorator do
 
     it 'returns population name when no species' do
       sd.object.content.update(:step01, name: 'pn')
-      expect(sd.label).to eq 'pn'
+      expect(sd.label).to eq '<span class="title">pn</span>'
     end
 
     it 'returns empty label when no name or species given' do
-      expect(sd.label).to eq I18n.t('submission.empty_label')
+      expect(sd.label).
+        to eq '<span class="title">' +
+              I18n.t('submission.empty_label') +
+              '</span>'
     end
   end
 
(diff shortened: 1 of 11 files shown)


Nuanda:
@kammerer Tomek, this is the last PR from yesterday - it's needed to have the front page more or less ready for styling and laying out. This changes pretty much in your code so a request for the review.

I actually start to doubt if having decorators rendering parts of templates is a good thing - we essentially hide presentation in classes. This may become more problematic when styling (not all markup is visible when looking at view/partial code).

spec/decorators/plant_population_submission_decorator_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    it 'returns empty label when no name or species given' do
+      expect(sd.label).
+        to eq '<span class="title">' +
+              I18n.t('submission.empty_label') +
+              '</span>'
+    end
+  end
+
+  describe '#further_details' do
+    it 'does not throw exception' do
```
kammerer: Does it add any value? In other test cases you call this method and if anything is thrown these cases will fail.

Nuanda: Yes it does (at least in my opinion) - essentially it makes sure the method is overloaded.

Nuanda: Oh ok, you mean - the other specs do test it as well. But they do that indirectly - so I prefer to have that kind of explicit test as well.

kammerer:
:+1: 

I usually do not make decorators output html but I am not strictly against it. I do not think it is much different from using helpers to do that. I am not sure if that is a problem for styling, if you use sensible editing tool you can quickly check the source anyway. Besides you use built-in helpers anyway.

Nuanda:
Well, I can easily change the decorator approach - actually I used it on your advice :).

When you look at the styling proposed by Bartosz you'll see he will need more detailed markup to introduce proper css - hence html in decorators. But I can drop that - I just didn't want to trespass on your code too much.

Nuanda:
Merging for now - we'll see how to refactor it when a new type of submission comes.

PR #86, opened 2015-03-28
Replaces PV and PVD with plant_line_details bound to plant_lines

### LINKED ISSUE #12: Triple repetition of species/genus in data
In CS data we have the species/genus combination of fields for:
 - plant_population
 - plant_line, which a plant_population is related to
 - plant_variety, which a plant_line is related to
 - (perhaps more, not detected now).

The issue is to sort out if this is on purpose or should we re-normalize it and, for instance, leave the species/genus data only in the plant_variety table. If that is the case, solve the problem of having data discrepancy between these tables.

Consult @sarahcayling and @wjurkowski wrt that.

This is the first in the, probably long, series of data cleansing tasks.

nowakowski:
Fixes #12

### DIFF
--- a/db/migrate/20150328155811_bind_plant_varieties_to_countries.rb
+++ b/db/migrate/20150328155811_bind_plant_varieties_to_countries.rb
@@ -62,10 +62,41 @@ class BindPlantVarietiesToCountries < ActiveRecord::Migration
 
     end
 
+    remove_column :plant_varieties, :country_of_origin
+    remove_column :plant_varieties, :country_registered
 
   end
 
   def down
+    add_column :plant_varieties, :country_of_origin, :string
+    add_column :plant_varieties, :country_registered, :string
+
+    pvs = ActiveRecord::Base.connection.execute("SELECT * FROM plant_varieties")
+    pvs.each do |pv|
+      pv_coo = ""
+      pv_cr = ""
+      coos = ActiveRecord::Base.connection.execute("SELECT country_code FROM \
+        plant_variety_country_of_origin WHERE plant_variety_name = \
+        E'#{escape(pv['plant_variety_name'])}'")
+      crs = ActiveRecord::Base.connection.execute("SELECT country_code FROM \
+        plant_variety_country_registered WHERE plant_variety_name = \
+        E'#{escape(pv['plant_variety_name'])}'")
+      pv_coo = (coos.collect {|coo| coo['country_code']}.join('; '))
+      pv_cr = (crs.collect {|cr| cr['country_code']}.join('; '))
+
+      unless pv_coo.blank?
+        update = "UPDATE plant_varieties SET country_of_origin = '#{pv_coo}' \
+          WHERE plant_variety_name = '#{escape(pv[plant_variety_name])}'"
+        ActiveRecord::Base.connection.execute(update)
+      end
+      unless pv_cr.blank?
+        update = "UPDATE plant_varieties SET country_registered = '#{pv_cr}' \
+          WHERE plant_variety_name = '#{escape(pv[plant_variety_name])}'"
+        ActiveRecord::Base.connection.execute(update)
+      end
+    end
+
+
     drop_table :plant_variety_country_of_origin
     drop_table :plant_variety_country_registered
     remove_index(:countries, :name => 'idx_ccs_country_code')
@@ -97,5 +128,12 @@ class BindPlantVarietiesToCountries < ActiveRecord::Migration
     code
   end
 
-
+  def escape(string)
+    if string.nil?
+      nil
+    else
+      pattern = /(\'|\"|\.|\*|\/|\-|\\|\)|\$|\+|\(|\^|\?|\!|\~|\`)/
+      string.gsub(pattern){|match|"\\"  + match}
+    end
+  end
 end
--- a/db/migrate/20150328155811_bind_plant_varieties_to_countries.rb
+++ b/db/migrate/20150328155811_bind_plant_varieties_to_countries.rb
@@ -31,7 +31,14 @@ class BindPlantVarietiesToCountries < ActiveRecord::Migration
             '#{code}')"
           ActiveRecord::Base.connection.execute(insert)
         else
-          puts "Unknown code #{code}. Giving up."
+          # Add notification to comments
+          puts "Unknown code #{code}. Inserting warning in comments."
+          comments = pv['comments']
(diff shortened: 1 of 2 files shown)


db/migrate/20150328144811_consolidate_plant_line_details.rb
```diff
@@ -0,0 +1,195 @@
...
+      t.string :year_registered,            default: "xxxx", null: false
+      t.string :breeders_variety_code
+      t.string :owner
+      t.string :quoted_parentage
+      t.string :female_parent
+      t.string :male_parent
+      t.string :detail_comments
+      t.string :detail_entered_by_whom,     default: "unspecified", null: false
+      t.date :detail_date_entered
+      t.string :detail_data_provenance
```
Nuanda: Since we merged two arrays I don't think keeping these detail_x fields is useful. Can we make a sensible "merge" of these values with the values from original PV fields?

db/migrate/20150328144811_consolidate_plant_line_details.rb
```diff
@@ -0,0 +1,195 @@
...
+
+    # Rename existing PV to avoid naming clash
+    ActiveRecord::Base.connection.execute("ALTER TABLE plant_varieties \
+      RENAME TO old_plant_varieties")
+
+    create_table :plant_varieties do |t|
+      t.string :plant_variety_name,         primary: true
+      t.string :crop_type
+      t.string :comments
+      t.text :entered_by_whom,              default: "unspecified", null: false
```
Nuanda: Can we drop that constraint, allow it to be null and remove the default?

db/migrate/20150328144811_consolidate_plant_line_details.rb
```diff
@@ -0,0 +1,195 @@
...
+      t.date :date_entered
+      t.string :data_provenance
+
+      t.string :data_attribution,           default: "unspecified", null: false
+      t.string :country_of_origin,          default: "xxx", null: false
+      t.string :country_registered,         default: "xxx", null: false
```
Nuanda: Can we make these two country_ fields explicit relations to the countries table?

db/migrate/20150328144811_consolidate_plant_line_details.rb
nowakowski: @Nuanda Can do that, although this will further upset the reverse migration (no way to restore missing data).

Nuanda: Don't worry, we still have the original dump, so nothing gets lost irrevocably.

Otherwise just imagine the UI form where we have both comments and detail_comments - it won't make any sense to the user, will it?

nowakowski: Indeed it won't. Proceeding as requested.

db/migrate/20150328144811_consolidate_plant_line_details.rb
nowakowski: By all means.

db/migrate/20150328144811_consolidate_plant_line_details.rb
nowakowski: 83 PVDs with country_of_origin not in countries.country_code. Most of them have multiple country codes in the field which calls for a m:n relationship. There are also three unidentified countries (XXX, CPB and PVD - @wjurkowski , can you explain these acronyms?) as well as a somewhat mysterious value of '<2'.

The `country_registered` field is a monumental mess which will require a smart curation task. I will get to work on the necessary transformations.

Nuanda: Well. Is countries table used anywhere else in the model? Perhaps we can drop it completely and just leave these two column values as a plain text (like they are now)?

nowakowski: I will try to sort this out; bear with me.

nowakowski:
OK, 'XXX' is simply shorthand for "missing data"; however I need help with the following codes:
- CAAS - the closest thing I've been able to find is the IATA code of the airport in Pahrump, Nevada, USA. Probably not what the authors had in mind. ;)
- CPB - Google search turns up a blank.
- PVD - same as above.
- GBBR - probably Great Britain (?)
- <2 - No idea. I suppose I'll just drop this value altogether.

nowakowski:
Two more:
- NPZ
- BRD

@wjurkowski - any suggestions?

Nuanda:
Piotr, don't worry about that. It's not that important - please go ahead with other tasks related to that part of the model, and leave countries as they are (for the time being, at least).

nowakowski:
Well, I've started working on it... seems a shame to give up halfway through; plus it's obstructing the PR - perhaps I can just ignore those 'alien' codes, assign everything else, create proper m:n relationships between `plant_varieties` and `countries` and leave those two fields in the table for future reference?

wjurkowski:
ja bym obstawial Bundes Republik Deutschland
Wyrzucil bym tez kody - jezeli nie jest jasno napisane. A ile NA w sumie?

2015-03-28 19:41 GMT+00:00 Piotr Nowakowski <notifications@github.com>:

nowakowski:
@wjurkowski Good catch. What do you mean by 'NA'?

nowakowski:
Assuming LIT stands for Lithuania (correct code is LTU).

nowakowski:
Also assuming the following:
- SPA is Spain (ESP)
- LAT is Latvia (LVA)
- GB is Great Britain (GBR)

nowakowski:
This leaves us with the following mysteries:
- NPZ
- CPB
- DAV
- CAAS
- KWS
- MOR (that's Mordor, right? ;)
- BSU

nowakowski:
We can review and merge as soon as @wjurkowski decides what to do with the "rogue states" listed above. :)

Nuanda:
Please add an index on countries.country_name (if not present) for fast search.

nowakowski:
@Nuanda Added.

wjurkowski:
should write NULL

2015-03-28 19:59 GMT+00:00 Piotr Nowakowski <notifications@github.com>:

nowakowski:
All done. The "new" `plant_varieties` model now correctly links to `countries` (via `pv.countries_of_origin` and `pv.countries_registered`). A warning message is added to the `comments` field whenever an unrecognized country appears in the input list.

Nuanda:
:+1:

PR #87, opened 2015-03-28
decomposed filter into two model concerns

Nuanda:
@kammerer Tomek, this is more or less what I was talking about. In principle:
 - Filterable to cover the filtering, both for :query and :search constraints, also takes care for strong_params, which are to be defined per-model
 - Pluckable (I know, a terrible name :)), that takes care for proper joins in multi-column pluck data retrieval.

So, your :search case needs only PlantLine.filter now, and I adjusted submissions.coffee to make proper query and parse the new result accordingly. NOTE - I'm not really sure where else I should change the :search query.

Anyway - including Filterable and Pluckable in subsequent models should make it much better to support further submissions and data tables, respectively.

Of course we still have "ugly" names, like table_columns or table_data - but I rather need that kind of specialized calls for smoother datatables coding.

I'm not merging it right now, tell me what you think first.

### DIFF
--- a/app/models/plant_population.rb
+++ b/app/models/plant_population.rb
@@ -21,28 +21,39 @@ class PlantPopulation < ActiveRecord::Base
                           foreign_key: 'plant_population_id',
                           association_foreign_key: 'plant_line_name'
 
-  # Exlude the ['none', 'unspecified', 'not applicable'] pseudo-record trio
-  scope :drop_dummies, -> do
-    where.not(canonical_population_name: '')
+  include Filterable
+
+  scope :by_name, -> { order(:plant_population_id) }
+
+  def self.grouped(params = nil)
+    count = 'count(plant_lines.plant_line_name)'
+    query = (params && params[:query].present?) ? filter(params) : all
+    query.
+      select(table_columns + [count]).
+      includes(:plant_lines).
+      group(table_columns).
+      by_name.
+      pluck(*(table_columns + [count]))
   end
 
-  def self.grouped(columns: nil, count: nil)
-    columns ||= [
+  private
+
+  def self.permitted_params
+    [
+      query: [
+        :plant_population_id
+      ]
+    ]
+  end
+
+  def self.table_columns
+    [
       'plant_populations.plant_population_id',
-      'plant_populations.species',
+      :species,
       :canonical_population_name,
       :female_parent_line,
       :male_parent_line,
       :population_type
     ]
-
-    count = 'count(plant_lines.plant_line_name)'
-
-    select(columns + [count]).
-      includes(:plant_lines).
-      group(columns).
-      drop_dummies.
-      order(:plant_population_id).
-      pluck(*(columns + [count]))
   end
 end
--- a/app/models/plant_population.rb
+++ b/app/models/plant_population.rb
@@ -29,7 +29,6 @@ class PlantPopulation < ActiveRecord::Base
     count = 'count(plant_lines.plant_line_name)'
     query = (params && params[:query].present?) ? filter(params) : all
     query.
-      select(table_columns + [count]).
       includes(:plant_lines).
       group(table_columns).
       by_name.
--- a/spec/controllers/plant_populations_controller_spec.rb
+++ b/spec/controllers/plant_populations_controller_spec.rb
@@ -2,19 +2,19 @@ require 'rails_helper'
 
 RSpec.describe PlantPopulationsController do
   context '#index' do
-    it 'returns table template on html request' do
+    it 'returns table template on html format request' do
       get :index
       expect(response).to render_template('plant_populations/index')
       expect(response).to render_template('layouts/application')
     end
 
-    it 'does not render htmls on ajax request' do
+    it 'does not render htmls on json format request' do
       get :index, format: :json
       expect(response).not_to render_template('plant_populations/index')
       expect(response).not_to render_template('layouts/application')
     end
 
-    it 'returns datatables json on ajax request' do
+    it 'returns datatables json on json format request' do
       pps = create_list(:plant_population, 2)
       get :index, format: :json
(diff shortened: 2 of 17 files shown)


Nuanda:
Well, we still have the problem of the same controller method responding to both queries. I retracted some coffee changes and now both ajax datatables get and ajax autocomplete get perform properly.

Nuanda:
Pushed the other half - for PP model and dT view.

spec/controllers/plant_populations_controller_spec.rb
```diff
@@ -23,5 +23,15 @@
       expect(json['data'].size).to eq 2
       expect(json['data'].map(&:first)).to match_array pps.map(&:id)
     end
+
+    it 'supports query filtering on ajax request' do
+      pps = create_list(:plant_population, 2).map(&:plant_population_id)
+      get :index, format: :json, query: { plant_population_id: pps[0] }
```
kammerer: Strictly speaking this is not ajax request (use `xhr :get, :index` if you want one). OTOH controller does not care about that, only about `Accept` header.

app/models/plant_population.rb
```diff
@@ -21,28 +21,39 @@ class PlantPopulation < ActiveRecord::Base
...
-    where.not(canonical_population_name: '')
+  include Filterable
+
+  scope :by_name, -> { order(:plant_population_id) }
+
+  def self.grouped(params = nil)
+    count = 'count(plant_lines.plant_line_name)'
+    query = (params && params[:query].present?) ? filter(params) : all
+    query.
+      select(table_columns + [count]).
```
kammerer: I just discovered that `select` is ignored when using `pluck`. E.g. `PlantPopulation.select("wrong_name").pluck(:plant_population_name)` -> `SELECT "plant_populations"."plant_population_name" FROM "plant_populations"`

Nuanda: Yes, I discovered that also. But after some fiddling with the code I decided to revert most of that I forgot about that. Will fix.

spec/controllers/plant_populations_controller_spec.rb
Nuanda: Good point - but the test is essentially proper, since it tests response on certain accepted content type. I'll change the wrong name.

kammerer: Sure, it was only about he name.

kammerer:
In general it is ok. I can live with `#table_data` in models. 
Is it a problem that one action is used for both autocomplete and datagrid?

Nuanda:
Well, it results in e.g. your submissions.coffee code being format-dependent (row[0]). Having proper json response it would be row.plant_line_name - but dataTables are really expecting tables, and changing that behavior would be more difficult.

But of course it's not the matter of model method name - it's rather the matter of the controller's index method. To solve it we would need params-based if switch there (if :search do one thing, if :query do the other) - which I tried and decided it to be uglier still ;).

So, no nicer solution for now. Just changed the model method to filtered, for better or worse.

kammerer:
I wasn't aware that tabular data was a requirement, I thought it is only optimization. Having JSON response for autocomplete would be more transparent indeed and I actually thought about implementing it via param switch. Not in the way you mentioned however, but by allowing to choose response format / used decorator (e.g. `http://...?as=table`).

Nuanda:
It looks like I could be wrong - that is datatables do ask for tabular data, but one may also reset it to work with json-encoded data. This of course required some extra js code (mapping between th elements and json attributes) but that's not that complex.

Then again, it may make the response somewhat bigger, but I'm not sure if that is of a concern to us.

Nuanda:
I tested for a rather larger (but not the largest) data table we have now - 3387 records (the datatable authors advise to keep client-side processing up to at least 10k records).

For array data we have 343 kB, for JSON data we have 2.3 MB - a lot more. This was for full-object JSON so I could cut some of it with SELECTing certain fields. However, what is worse is the js processing (accessing object fields instead of array records) - it considerably freezes the view.

So, I think this is a too large price to pay for having nicer coffee code her or there :). Let's keep it as it is now. Also - we only have that problem with the tables we want to run autocomplete on (still, in the end of the day, there could be quite a few of them).

kammerer:
We should definitely stay with array data for tables, and I think we can
leave it like that for autocomplete selects too (unless sth changes and we
will need json).

2015-03-31 16:42 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> I tested for a rather larger (but not the largest) data table we have now
> - 3387 records (the datatable authors advise to keep client-side processing
> up to at least 10k records).
>
> For array data we have 343 kB, for JSON data we have 2.3 MB - a lot more.
> This was for full-object JSON so I could cut some of it with SELECTing
> certain fields. However, what is worse is the js processing (accessing
> object fields instead of array records) - it considerably freezes the view.
>
> So, I think this is a too large price to pay for having nicer coffee code
> her or there :). Let's keep it as it is now. Also - we only have that
> problem with the tables we want to run autocomplete on (still, in the end
> of the day, there could be quite a few of them).
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/87#issuecomment-88115249>.
>



-- 
Tomasz Szymczyszyn

Nuanda:
There is yet another solution. I suspect having a lot of <DataModel>sController classes with similar single index operation with almost same content. I'm thinking about having a single generic DataTablesController with single, generic, index function, to server parameter-set model data (what models could be queried that way I may set in strong parameters of this generic controller).

If I do that I'm going to have a similar-names function in all datatableable ;) models, so it's probably table_data again.

But then we might keep the <DataModel>sController classes for the models where we need something besides data table query, e.g. an autocomplete. The index operation of these controllers will serve your search query and we may drop decorators there and we also may have plain json returned. You may still call Filterable#filter method/scope for that (as I'm going to call it in my other scopes).

What do you think?

PR #88, opened 2015-03-28
Extends PP model and curation task to bind all PPs to TTs

### LINKED ISSUE #19: Checking plant_population - plant_line genus and species consistency
Next step after #12 and #16 is to get rid of, if possible, columns _genus_ and _species_ from the plant_populations (PP) table. There are three issues connected to that.

1. **No PLs**. As @nowakowski found out, there are 55 PPs that have no assigned plant_line (PL). This is the "commercial line" problem already discussed with @wjurkowski via email.

2. **DIfferent genus**. There are 2 PPs defined as Brassica yet they are linked to 1+ PLs that are set as X_Brassicoraphanus.

3. **Different species**. There are 5 PPs that are not defined as Brassica spp, yet they link to various species of PLs. These PPs are: BolDFS_00, BolDFS_01, BolDFS_04, BolDCC_02, BolEC03_01 (btw all of them are defined as Brassica oleracea, but this is probably a coincidence).

In my opinion, after removal of the genus/species columns from PPs, we should state:
 - unspecified, in case 1
 - various, in case 2
 - Brassica spp, in case 3.

Is that a correct way out?

nowakowski:
Fixes #19

### DIFF
--- a/app/models/plant_population.rb
+++ b/app/models/plant_population.rb
@@ -1,5 +1,7 @@
 class PlantPopulation < ActiveRecord::Base
 
+  belongs_to :taxonomy_term
+
   belongs_to :population_type_lookup, foreign_key: 'population_type'
 
   belongs_to :male_parent_line, class_name: 'PlantLine',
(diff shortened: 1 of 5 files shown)


Nuanda:
:+1:

PR #94, opened 2015-03-29
fixes due to PP switch from genus and species to tt relation

### LINKED ISSUE #93: Add taxonomy term selection in the 2nd step of PP submission
Don't forget to check plant_population_submission_decorator and its corresponding spec.

Nuanda:
Fixes #93 

@kammerer This _almost_ works. I was able to read through a lot of the submissions magic, but still something somewhere prevents me from setting TT in the 2nd step. I saw the chosen TT is not submitted to the update method within :content params hash, but it's almost 1:30 am and I'm really not able to decipher submissions any further :). So I need your help to finish this.

Gosh, that was a lot of changes for something so silly.

@nowakowski Remember, you need to remove columns from the initial "big" migration as well. Otherwise test env is different to the dev/prod env, if you just remove columns in rake tasks.

### DIFF
--- a/app/decorators/plant_population_submission_decorator.rb
+++ b/app/decorators/plant_population_submission_decorator.rb
@@ -6,7 +6,7 @@ class PlantPopulationSubmissionDecorator < SubmissionDecorator
       :span,
       ''.tap do |l|
         l << population_name if population_name
-        l << "(#{species_name})" if species_name
+        l << " (#{species_name})" if species_name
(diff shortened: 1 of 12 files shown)


Nuanda:
Looks ok now :) - thanks, I would never figure that out on my own. Any further comments regarding my code?

kammerer:
No, it is ok :+1:

PR #95, opened 2015-03-29
Add rake task encapsulating all db preparation steps

kammerer:
Provide a Rake task performing everything which is required to prepare database (as described in wiki)

### DIFF
--- a/lib/tasks/app.rake
+++ b/lib/tasks/app.rake
@@ -1,5 +1,5 @@
-namespace :db do
-  desc "Prepare DB from zero to production-ready state"
+namespace :app do
+  desc "Prepare DB and ES from zero to production-ready state"
   task :bootstrap => :environment do
     unless ask_for_confirmation
       puts "Exiting."
@@ -9,6 +9,8 @@ namespace :db do
     recreate_db
     puts " - loading cropstore DB dump data"
     restore_cropstore_dump
+    puts " - building ES indices"
+    build_elasticsearch_indices
     puts " - initial curation of CS data"
     perform_initial_curation
     puts " - adding new BIP migrations"
@@ -38,6 +40,10 @@ namespace :db do
     unless system(env, cmd)
       raise "Cropstore dump restoration failed: #{$?}"
     end
+  end
+
+  def build_elasticsearch_indices
+    Rake::Task['elasticsearch:import:all'].invoke
   end
 
   def perform_initial_curation
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -5,12 +5,19 @@ namespace :db do
       puts "Exiting."
       next
     end
+    puts " - dropping old DB and creating a new one"
     recreate_db
+    puts " - loading cropstore DB dump data"
     restore_cropstore_dump
+    puts " - initial curation of CS data"
     perform_initial_curation
+    puts " - adding new BIP migrations"
     migrate_db
+    puts " - loading Gramene and CS taxonomy"
     bootstrap_obo_taxonomy
+    puts " - final curation of CS/taxonomy data"
     perform_final_curation
+    puts " - DONE"
   end
 
   def ask_for_confirmation
@@ -34,9 +41,11 @@ namespace :db do
   end
 
   def perform_initial_curation
-    Rake::Task['curate:purge_empty_tables'].invoke
-    Rake::Task['curate:purge_empty_columns'].invoke
-    Rake::Task['curate:purge_placeholder_records'].invoke
+    silence_stdout do
+      Rake::Task['curate:purge_empty_tables'].invoke
+      Rake::Task['curate:purge_empty_columns'].invoke
+      Rake::Task['curate:purge_placeholder_records'].invoke
+    end
   end
 
   def migrate_db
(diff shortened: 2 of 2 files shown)


kammerer:
BTW When I prepare db afresh and export schema it is different that the schema commited to master.

Nuanda:
I think we should commit the "clear" schema - that is the one built by running all migrations, e.g. for the test environment. Otherwise Travis, which relies on that schema, will not pass our tests. So only commit to master "clean" schema from all migrations, not what results of our DB-preparation process.

kammerer:
I do no get it. Why should we have different schemas for tests and for development / production?

Nuanda:
Well, it'd not that we should - but we do. One "schema" is created by importing a ready DB and then migrating the rest of migrations. The other schema is by importing nothing and running all migrations. The fact of life is that the initial "big" migration, that @nowakowski provided to serve as DB dump import representation for testing, is perhaps slightly different from what we have in the real data, right?

kammerer:
Ok, I think I understand now, but doesn't that mean that the "clean" schema
is without much value since it is further from real data that we will
eventually need to import in production?

2015-03-29 22:47 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Well, it'd not that we should - but we do. One "schema" is created by
> importing a ready DB and then migrating the rest of migrations. The other
> schema is by importing nothing and running all migrations. The fact of life
> is that the initial "big" migration, that @nowakowski
> <https://github.com/nowakowski> provided to serve as DB dump import
> representation for testing, is perhaps slightly different from what we have
> in the real data, right?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/95#issuecomment-87473989>.
>



-- 
Tomasz Szymczyszyn

kammerer:
BTW schema issue was only a side note, this PR actually does not change
schema.

2015-03-29 22:54 GMT+02:00 Tomek Szymczyszyn <tomasz.szymczyszyn@gmail.com>:

> Ok, I think I understand now, but doesn't that mean that the "clean"
> schema is without much value since it is further from real data that we
> will eventually need to import in production?
>
> 2015-03-29 22:47 GMT+02:00 Tomasz Gubała <notifications@github.com>:
>
>> Well, it'd not that we should - but we do. One "schema" is created by
>> importing a ready DB and then migrating the rest of migrations. The other
>> schema is by importing nothing and running all migrations. The fact of life
>> is that the initial "big" migration, that @nowakowski
>> <https://github.com/nowakowski> provided to serve as DB dump import
>> representation for testing, is perhaps slightly different from what we have
>> in the real data, right?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/eSpectrum-IT/brassica/pull/95#issuecomment-87473989>.
>>
>
>
>
> --
> Tomasz Szymczyszyn
>



-- 
Tomasz Szymczyszyn

Nuanda:
That we have different schemas is essentially the consequence of the decision we have taken on how to deal with the CS data dump vs. rails migrations. I proposed to have clear migrations from the start and then to prepare elaborate rake task to import (seed) the DB from the CS dump to the current rails model.

However @nowakowski opted for the solution we have now and, though I kind of regret it, "it is what it is". Also this creates problems like yesterday's @nowakowski PR, which changed schema in a rake task but didn't do it in the initial "big" schema - so we have to be _really careful_ now to try to keep the migrations' schema as close to the dump (dev/prod) schema as possible. Otherwise our tests will "fool" us staying all green, but against an outdated schema :(...

Nuanda:
Is this PR ready for review?

kammerer:
Shouldn't we just assume we work with schema originating from the dump and
commit it? It seems to me that dump is the real initial state and first
migration is just an imperfect imitation ;)

2015-03-29 23:03 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> That we have different schemas is essentially the consequence of the
> decision we have taken on how to deal with the CS data dump vs. rails
> migrations. I proposed to have clear migrations from the start and then to
> prepare elaborate rake task to import (seed) the DB from the CS dump to the
> current rails model.
>
> However @nowakowski <https://github.com/nowakowski> opted for the
> solution we have now and, though I kind of regret it, "it is what it is".
> Also this creates problems like yesterday's @nowakowski
> <https://github.com/nowakowski> PR, which changed schema in a rake task
> but didn't do it in the initial "big" schema - so we have to be *really
> careful* now to try to keep the migrations' schema as close to the dump
> (dev/prod) schema as possible. Otherwise our tests will "fool" us staying
> all green, but against an outdated schema :(...
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/95#issuecomment-87475662>.
>



-- 
Tomasz Szymczyszyn

kammerer:
Yes, it is ready. Forgot to add label, sorry.

Nuanda:
This is not so simple - do you want to run all tests (locally) on the actual CS DB dump with all the records inside? Remember that local rspec does not use schema.rb and relies on what you migrate your test db to.

Nuanda:
Ok, let's move the schema/dump conversation to email - if you have some idea how to better deal with that... situation, write me and @nowakowski an email.

Going back to the actual PR. I'm thinking about moving all rake tasks outputs to a dedicated db_bootstrap_log and only showing some simple messages (e.g. what step was performed right at a given moment) on STDOUT. Is it possible to invoke a rake task and redirect all output to a file?

Nuanda:
E.g. sth like that:

def silence_stdout
  $stdout = File.new( '/dev/null', 'w' )
  yield
ensure
  $stdout = STDOUT
end

Nuanda:
Pushed stdout silencer. :+1:

Nuanda:
Before you merge, perhaps you could add that rake task in the process, probably somewhere before the plant_taxonomy rake tasks:
rake environment elasticsearch:import:all

kammerer:
Added ES indexing step.

Achtung! I changed task name from db:bootstrap to app:bootstrap.

PR #96, opened 2015-03-29
Setup full text search

### LINKED ISSUE #72: Setup full text search solution based on elastic search
For starters, mine data in PPs and PLs.

kammerer:
PL and PP indexing. Right now the indices are pretty simplistic, we probably should differentiate weights of different fields to make results more useful.

Fixes #72


Nuanda:
:+1:

PR #104, opened 2015-04-01
test commit, sass files restructured, haml structure changes

bchucherko:
I created a new branch.
Some test changes added:
- sass files restructured for easier management
- some changes in layouts


db/schema.rb
```diff
@@ -11,12 +11,78 @@
 #
```
kammerer: Please, do not commit db/schema.rb

app/assets/stylesheets/vendor/_select2.sass
```diff
@@ -0,0 +1,422 @@
+.select2-container
```
kammerer: This one should go to vendor/assets/stylesheets/select2.sass. I put it in wrong place. Additionally app/assets/stylesheets/select2.css needs to be removed if we have .sass version.

kammerer:
Two minor issues to fix but overall :+1:

app/assets/stylesheets/vendor/_select2.sass
bchucherko: OK.
I don't know where to find <head> of app where css files (bootstrap, application, select) are loaded.
Furthermore I'm not sure how sass compilation works right now - it seems only application.sass is compiled - thus also this file (select3.sass) would have to be added to compilation as well or it can be compiled manually once (if it's more like a lib).

kammerer: This is default stylesheet for a JS lib, that's why it could go to
vendor/assets. Overrides (if any) should be placed in app/assets.

Application.sass is the "root" or so called manifest file for asset
pipeline. There are two ways to include other files - via @import (which
must be used for sass/scss) or via comment (e.g. *= dataTables.bootstrap).
The other way can be used for css but must not be used for sass/scss, so it
actually shouldn't be used for select2 right now (I missed that one in PR).



2015-04-02 8:08 GMT+02:00 bchucherko <notifications@github.com>:

> In app/assets/stylesheets/vendor/_select2.sass
> <https://github.com/eSpectrum-IT/brassica/pull/104#discussion_r27634334>:
>
> > @@ -0,0 +1,422 @@
> > +.select2-container
>
> OK.
> I don't know where to find
> of app where css files (bootstrap, application, select) are loaded.
> Furthermore I'm not sure how sass compilation works right now - it seems
> only application.sass is compiled - thus also this file (select3.sass)
> would have to be added to compilation as well or it can be compiled
> manually once (if it's more like a lib).
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/104/files#r27634334>.
>



-- 
Tomasz Szymczyszyn

Nuanda:
@bchucherko Don't mind the schema.rb comment - I'll take care of that when merging. Please commit two other changes @kammerer requested and I'll merge this PR.

app/assets/stylesheets/vendor/_select2.sass
Nuanda: Should we move the new _select2.sass to the vendor/assets or is it better to move there the original select2.css file?

Nuanda:
Merged that since it was getting too old (and would eventually grow even more problematic for future merge). Filed another issue #131 to solve the select2 stylesheet problem.

app/assets/stylesheets/vendor/_select2.sass
Nuanda: bump

PR #107, opened 2015-04-02
Green table ids

nowakowski:
This PR replaces all PKs and FKs in the "green" portion of the CropstoreDB model with autoincremented numerical keys. Indices have been added to DB columns where appropriate. New model classes are provided. Rspecs have been updated. A new database dump has been generated and added to the app:bootstrap task.


Nuanda:
:+1:

lib/migration_helper.rb
```diff
@@ -0,0 +1,112 @@
+# This file aggregates helper methods which are used in various migrations
+module MigrationHelper
+
+  # Escapes a string literal so that it can be safely fed into PGSQL queries
+  # NOTE: with literals of this type prefix the string delimiters with E, e.g.:
+  # SELECT * FROM plant_populations WHERE plant_population_name = E'#{escaped_name}';
+  # (Because SQL standards are for losers... ;)
+  def escape(string)
```
kammerer: Maybe `ActiveRecord::Base.connection.quote(string)` would be useful?

nowakowski: Deprecated in favor of `quote_value` but thanks all the same - seems no one on stackoverflow noticed it... ;)

kammerer: You can score some points now ;)

2015-04-02 21:27 GMT+02:00 Piotr Nowakowski <notifications@github.com>:

> In lib/migration_helper.rb
> <https://github.com/eSpectrum-IT/brassica/pull/107#discussion_r27688839>:
>
> > @@ -0,0 +1,112 @@
> > +# This file aggregates helper methods which are used in various migrations
> > +module MigrationHelper
> > +
> > +  # Escapes a string literal so that it can be safely fed into PGSQL queries
> > +  # NOTE: with literals of this type prefix the string delimiters with E, e.g.:
> > +  # SELECT * FROM plant_populations WHERE plant_population_name = E'#{escaped_name}';
> > +  # (Because SQL standards are for losers... ;)
> > +  def escape(string)
>
> Deprecated in favor of quote_value but thanks all the same - seems no one
> on stackoverflow noticed it... ;)
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/107/files#r27688839>.
>



-- 
Tomasz Szymczyszyn

PR #108, opened 2015-04-02
big data tables handling refactoring

Nuanda:
A lot of things in here:
 - data tables call DataTablesController now, passing required model as a param (values filtered through dedicated controller method)
 - `ModelClass`sControllers index methods may serve only automplete now, not bothering about pluck, item ordering and such (calling Filterable.filter directly)
 - added two new models for datatables: PlantTrials and TraitDescriptors, refactored PL and PP models to fit the new querying strategy
 - Model.table_columns is now exposed for the view to use it to render `th` elements.

One possible issue: PL.filter will not return autocomplete results in any particular order - is it important or was it just a by-product of using previous model methods?

The views/data_tables/`model`_table views may be generalized even further (up to the point of a single view template, perhaps) but I didn't want to do that now, in this already large PR.

This is more or less what I proposed in #87 discussion.


app/views/data_tables/plant_lines_table.html.haml
```diff
@@ -0,0 +1,10 @@
+= render 'data_tables/tabs'
+
+.lead Plant Lines table
+
+= link_to 'TEMP LINK TO Plant populations', data_tables_path(model: :plant_populations)
+
+= datatable_tag do
+  - PlantLine.table_columns.each do |table_column|
```
kammerer: This part could be extracted to a helper to avoid repetition.

app/helpers/application_helper.rb
```diff
@@ -28,10 +28,15 @@ def active_link(label)
...
-      [:browse, plant_populations_path],
+      [:browse, browse_data_path],
       [:submit, new_submission_path],
       [:api, '#'],
       [:about, about_path]
     ]
   end
+
+  # Use this as the default main path for browsing UI section
+  def browse_data_path
```
kammerer: Why not define it in config/routes.rb?

kammerer:
:+1: I like it, cannot test though since app:bootstrap suddenly cased to work :( Probably local issue since it fails both with old and new dump.

app/helpers/application_helper.rb
Nuanda: Will try to do it.

app/views/data_tables/plant_lines_table.html.haml
Nuanda: I know, but as I mentioned in the PR description, I deliberately don't do it (yet).

Nuanda:
Well, not worth testing it yet since now I need to merge it with the other enormous PR of Piotr. This is going to be... fun?

app/views/data_tables/plant_lines_table.html.haml
kammerer: Right, it was such a long read that I forgot you mentioned it ;)

2015-04-02 21:02 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> In app/views/data_tables/plant_lines_table.html.haml
> <https://github.com/eSpectrum-IT/brassica/pull/108#discussion_r27686944>:
>
> > @@ -0,0 +1,10 @@
> > += render 'data_tables/tabs'
> > +
> > +.lead Plant Lines table
> > +
> > += link_to 'TEMP LINK TO Plant populations', data_tables_path(model: :plant_populations)
> > +
> > += datatable_tag do
> > +  - PlantLine.table_columns.each do |table_column|
>
> I know, but as I mentioned in the PR description, I deliberately don't do
> it (yet).
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/108/files#r27686944>.
>



-- 
Tomasz Szymczyszyn

PR #116, opened 2015-04-03
New PKs/FKs in the 'red' portion of the CropstoreDB data model

nowakowski:
Database refactoring covering the 'red' portion of the CropstoreDB model - new primary/foreign keys and model classes.


Nuanda:
Will my query in TraitDescriptor still work?

Can you run a local server, go to Browse data, switch to the second Traits tab and see if data is loaded (wait for ca 12 secs)?

nowakowski:
@Nuanda You'll be happy to know it still works. :)

Nuanda:
Did you add indices in Critical Places(tm)?

Also, how about removing old (perhaps apart from the very first) db dumps, that are not useful anymore?

nowakowski:
@Nuanda That depends on what you consider a Critical Place(tm); however my helper automatically adds indices on all FKs it creates (PKs get indices for free from the DBMS).

We will probably need to add more indices as it becomes clear which model fields are frequently used in queries.

Regarding old dump - I thought I had done just that (cf. the contents of `db/`). Is there any specific dump you would like me to delete?

Nuanda:
:+1:

nowakowski:
@Nuanda @kammerer Don't forget to reload your DB from the new dump (or run the new migration).

PR #117, opened 2015-04-03
simple counter solution for long query

Nuanda:
Makes the query both fast and simple. Provided the data submission process would be infrequent, perhaps having a simple counter is all right.


lib/tasks/app.rake
```diff
@@ -28,12 +30,16 @@ namespace :app do
   def restore_cropstore_dump
     db_config = Rails.application.config_for(:database)
     env = { "PGUSER" => db_config["username"], "PGPASSWORD" => db_config["password"] }
-    cmd = "pg_restore -O -d #{db_config["database"]} db/cropstore_web_20150403_1530.dump"
+    cmd = "pg_restore -O -d #{db_config["database"]} db/cropstore_dev_20150403_1530.dump"
```
nowakowski: Please undo this change; I've already fixed the filename in master.

app/models/plant_population.rb
```diff
@@ -1,6 +1,6 @@
 class PlantPopulation < ActiveRecord::Base
-  include Elasticsearch::Model
-  include Elasticsearch::Model::Callbacks
+  # include Elasticsearch::Model
+  # include Elasticsearch::Model::Callbacks
```
nowakowski: Why are these commented out?

Nuanda: fixed

lib/tasks/app.rake
Nuanda: So it should merge properly.

nowakowski: I renamed the file itself, not the reference in `app.rake`. It won't merge properly.

nowakowski:
:+1:

PR #124, opened 2015-04-08
Plant varieties table

Nuanda:
Includes considerable changes on both model and view layers. Piotr, I would appreciate if you review this (also for the sake of learning a bit more about non-model layers of BIP :)).


app/models/plant_population.rb
```diff
@@ -32,22 +32,22 @@ class PlantPopulation < ActiveRecord::Base
...
   def self.table_columns
     [
       'plant_populations.name',
       'taxonomy_terms.name',
-      :canonical_population_name,
+      'canonical_population_name',
       'plant_lines.plant_line_name',
       'male_parent_lines_plant_populations.plant_line_name',
-      'pop_type_lookup.population_type'
+      'pop_type_lookup.population_type',
```
nowakowski: Won't this fail for the 'problematic' population which has no associated `pop_type_lookup` record?

Nuanda: Which population is that?

Nuanda: I checked, it does work.

nowakowski:
Curious: even though `data_tables_controller` whitelists :model in query params, I'm still getting the following message in my local client:

`Started GET "/data_tables?model=plant_lines&query%5Bid%5D=15162&_=1428590088167" for 127.0.0.1 at 2015-04-09 16:34:48 +0200
Processing by DataTablesController#index as JSON
  Parameters: {"model"=>"plant_lines", "query"=>{"id"=>"15162"}, "_"=>"1428590088167"}
Unpermitted parameters: model, _
Completed 200 OK in 3ms (Views: 0.3ms | ActiveRecord: 0.8ms | Elasticsearch: 0.0ms)`

..still the request appears to complete successfully. Can you suggest what might be wrong?

Nuanda:
Nothings wrong. The message is from PlantLines.permitted_params. During the second, Ajax call to the DataTablesController.index, it queries the model and the given model excludes everything apart from allowed query. Yet the params[:model] IS needed just before that - in the DataTablesController.index method (so it knows which model to call). It is simply filtered out later on which is perfectly ok.

Nuanda:
BTW it's the Filterable model concern that takes care for that parameters filtering.

nowakowski:
@Nuanda Thanks for explaining. :) :+1:

PR #128, opened 2015-04-09
Drops obsolete NOT NULL constraints.

### LINKED ISSUE #36: Do not require things like comments, description, data_provenance
I know the original CS author always required values for these fields, and then supplied some placeholder like 'unspecified' if the user did not provide any value.

However, it is probably much better to not require these values in the first place and allow NULLs in the database.

nowakowski:
Fixes #36


nowakowski:
Placeholder PR; do not merge.

nowakowski:
Some additional NNs may be dropped depending on the outcome of #133

Nuanda:
:+1:

PR #129, opened 2015-04-10
Annotations concern

### LINKED ISSUE #125: Generic solution for the five common model fields
We have a multitude of tables where these fields repeat:
-    t.text "comments"
-    t.text "entered_by_whom"
-    t.date "date_entered"
-    t.text "data_provenance"
-    t.text "data_owned_by"

Find a good strategy to handle these fields for table views.

Nuanda:
Fixes #125 

For every model mentioned there (i.e. each one that contains those 4-5 generic columns):
 - the table data call is extended by a single (always last) column with the record id
 - that id is used to construct a link to data_tables_controller#show method
 - the method returns a json with those 4-5 field values
 - that json is fed into a simple popover.

This PR _does not_ provide a finished UI experience for this - it just leaves it in the stage that the data transfer is clearly visible yet the result is not presented to the user in any satisfactory way.

### DIFF
--- a/app/assets/javascripts/data_grid.js.coffee
+++ b/app/assets/javascripts/data_grid.js.coffee
@@ -37,7 +37,41 @@ $ ->
 
 
   $('.data-table').each (i)->
-    $(this).DataTable window.configs[this.id]
+    specificConfig = window.configs[this.id]
+    if specificConfig
+      specificConfig['columnDefs'] =
+        $.merge(specificConfig['columnDefs'], window.baseColumnDefs(this.id))
+    else
+      specificConfig =
+        columnDefs: window.baseColumnDefs(this.id)
+    $(this).DataTable specificConfig
+
+
+  $('body').popover
+    selector: '[data-popover-source]'
+    placement: 'left'
+    title: 'Record metadata'
+    trigger: 'focus'
+    html: 'true'
+    content: ->
+      divId = "tmp-id-" + $.now()
+      $.ajax
+        url: $(this).data('popover-source')
+        success: (response) ->
+          $('#'+divId).html(JSON.stringify(response))
+      "<div id='#{divId}'>Loading...</div>"
+
+
+window.baseColumnDefs = (model) ->
+  [
+    targets: 'annotations'
+    render: (data, type, full, meta) ->
+      objectId = full[full.length - 1]
+      if objectId
+        '<button data-popover-source="data_tables/' + objectId + '?model=' + model + '">Metadata</button>'
+      else
+        ''
+  ]
 
 
 # Specific configurations for particular DataTables, including callbacks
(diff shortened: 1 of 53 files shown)


Nuanda:
@nowakowski Piotr, since this is heavy on the model-side, I'd like to ask you for the review.

app/models/map_position.rb
```diff
@@ -6,4 +6,5 @@ class MapPosition < ActiveRecord::Base
 
   has_many :map_locus_hits, foreign_key: 'map_position'
 
+  include Annotable
```
nowakowski: This might fail since `map_positions` currently does not have a primary key (or an `id` field).

nowakowski: Oops, I just remembered I had added this field yesterday. Disregard the previous comment.

nowakowski:
:+1:

app/assets/javascripts/data_grid.js.coffee
```diff
@@ -73,22 +107,22 @@ window.configs =
       [
         targets: [3]
```
kammerer: This numbers look cryptic. What do you think about aliasing them? Of course JS will not know if these aliases are correct but at least they will hint what was intended.

Nuanda: Would it be better if I attach a column_name_column html class to each column and then change that js declaration into: `tagets: 'column_name_column'` ?

spec/factories/plant_line.rb
```diff
@@ -3,10 +3,7 @@
     sequence(:plant_line_name) {|n| "#{Faker::Lorem.characters(5)}_#{n}"}
     common_name { Faker::Lorem.word }
     previous_line_name { Faker::Lorem.word }
-    date_entered { Faker::Date.backward }
-    data_owned_by { Faker::Company.name }
-    comments { Faker::Lorem.sentence }
-    data_provenance { Faker::Lorem.sentence }
     taxonomy_term
+    instance_eval &AnnotableFactory.annotated
```
kammerer: Not saying this is bad solution, but the same should be possible with a `trait`:

```ruby
FactoryGirl.define do
  trait :annotable do
     date_entered { ... }
  end

  factory :plant_line do
    annotable
  end
end
```

Nuanda: Can you define such a trait in a common file and then use it in multiple factories?

app/assets/javascripts/data_grid.js.coffee
Nuanda: Pushed this new solution to #143 - see it there.

kammerer: Looks good.

spec/factories/plant_line.rb
kammerer: Yup

Nuanda: Done.

PR #135, opened 2015-04-11
Numerical PKs/FKs in the blue portion of the CropstoreDB model.

### LINKED ISSUE #68: We need to do something about restriction_enzymes
In the CropstoreDB model this class links to `marker_assays` (it's actually a double relationship - via `restriction_enzyme_A` and `restriction_enzyme_B`), however these two columns had no meaningful FK values and were therefore dropped by our data curation scripts. At present, the `restriction_enzymes` table is entirely detached from the data model. We may restore the relationship with `marker_assays` but the question arises: how can existing records in that table be linked to records in `restriction_enzymes`?

Comments are welcome.

nowakowski:
Completes the basic PK/FK refactoring of the CropstoreDB model. Also fixes #68.


app/models/marker_assay.rb
```diff
@@ -1,21 +1,20 @@
...
   belongs_to :primer_B, class_name: 'Primer',
-             foreign_key: 'primer_b'
+             foreign_key: 'primer_b_id'
 
-  belongs_to :probe, foreign_key: 'probe_name'
+  belongs_to :probe
 
-  has_many :population_loci, class_name: 'PopulationLocus',
-           foreign_key: 'marker_assay_name'
+  has_many :population_loci, class_name: 'PopulationLocus'
```
Nuanda: I have added an inflection initializer in my latest PR, solving the loci - locus problem. Please check if that class_name declaration is still needed.

app/models/population_locus.rb
```diff
@@ -2,7 +2,7 @@ class PopulationLocus < ActiveRecord::Base
   self.table_name = 'population_loci'
```
Nuanda: Again, this _might_ not be needed anymore - please test.

app/models/primer.rb
```diff
@@ -1,9 +1,9 @@
 class Primer < ActiveRecord::Base
 
   has_many :marker_assays_A, class_name: 'MarkerAssay',
```
Nuanda: Is that capital A/B important? Since it looks so un-rdbms-y ;). But if that is required...

db/schema.rb
```diff
@@ -156,23 +156,39 @@
...
+  add_index "marker_assays", ["marker_sequence_assignment_id"], name: "marker_assays_marker_sequence_assignment_id_idx", using: :btree
+  add_index "marker_assays", ["primer_a_id"], name: "marker_assays_primer_a_id_idx", using: :btree
+  add_index "marker_assays", ["primer_a_name"], name: "marker_assays_primer_a_idx", using: :btree
+  add_index "marker_assays", ["primer_b_id"], name: "marker_assays_primer_b_id_idx", using: :btree
+  add_index "marker_assays", ["primer_b_name"], name: "marker_assays_primer_b_idx", using: :btree
+  add_index "marker_assays", ["probe_id"], name: "marker_assays_probe_id_idx", using: :btree
+
+  create_table "marker_sequence_assignments", force: :cascade do |t|
     t.text "marker_set",              default: "unspecified", null: false
+    t.text "canonical_marker_name",   default: "",            null: false
```
Nuanda: Does it have anything in common with MarkerAssay.canonical_marker_name?

app/models/marker_assay.rb
nowakowski: You're right; it isn't. Removed.

app/models/population_locus.rb
nowakowski: Yup.

app/models/primer.rb
nowakowski: That's because `marker_assays` links to `primers` twice (via `primer_a_id` and `primer_b_id`). I've added a `marker_assays` method to `Primer` which pulls a union of both lists.

Nuanda: No no, I am asking if we can change it to _a and _b instead of _A and _B, which seems awkward inside a DB.

db/schema.rb
nowakowski: It's the old PK of `marker_sequence_assignments`. `canonical_marker_name` was left as a text column in `marker_assays` for the usual reason (lots of unmatched records - removing this column would result in loss of data).

app/models/primer.rb
```diff
@@ -1,9 +1,13 @@
...
-             foreign_key: 'primer_a'
+             foreign_key: 'primer_a_id'
   has_many :marker_assays_B, class_name: 'MarkerAssay',
-           foreign_key: 'primer_b'
+             foreign_key: 'primer_b_id'
+
+  def marker_assays
+    marker_assays_A | marker_assays_B
```
Nuanda: What does that do?

app/models/primer.rb
nowakowski: Oh, sure.

app/models/primer.rb
nowakowski: See my response to your previous comment (in `primer.rb`). I can delete it if you wish.

Nuanda: Does this method run two queries and join the results ruby-side or does it combine them in the DB. Asking out of pure curiosity.

Nuanda:
Anyway, regardless the last question, it's :+1:

Issue #136, opened 2015-04-11
Help needed deciphering CS QTL query


Nuanda:
Piotr, I need your help with one query. Please see the QTL tab here http://www.interstoredb.org/Brassica/cs/. Then please check the CS_create_aux_tables_Queries_update_16_06_2014.txt file inside our dropbox folder and see the query responsible for rendering that table.

Can you explain to me what is exactly shown in the column second to last (Count scores)? It looks like it depends on some other temporary table/DB view...

nowakowski:
The query retrieves the value of a column called `count_score` - this is not defined in any of the standard model tables in the WHERE clause so it must be defined in `auxTraitsInterface` but I cannot find the definition of that entity anywhere in the file (or indeed in the project tree). Did you define such an entity (and if so - where)?

Perhaps you wanted `td.trait_scores_count` instead?

Nuanda:
Me? How could _I_ define it as we are talking the original CS queries here? :)

nowakowski:
Oh. Of course! Something is missing from the model / was not exported along with the database. I'll check if it's my problem or TGAC's problem...

nowakowski:
No, it's not one of mine. The original MySQL data dump does not include any such table/view/stored procedure. We need to ask @wjurkowski to take a look at this.

Nuanda:
Could it be it _somehow_ refers to the 'CREATE TABLE aux_trait_interface' definition (also present in the same file)?

nowakowski:
It's possible (but only possible!) - that table defines something like this:
`count(score_value) AS 'Count Scores'
...in which case this would simply be the number of `trait_scores` records retrieved for a given `trait_descriptor`.

Nuanda:
That's what I thought and that's what I tried to do in the code - see #137, app/models/qtl.rb file. However, I get different numbers...

Perhaps the counter itself, located in the TraitDescriptor (called `trait_scores_count`) is wrong? I don't know.

Can you see this PR and test, what's wrong with my query?

nowakowski:
The core of the problem is that there are multiple `trait_descriptors` with identical names.

The CropstoreDB application does something silly to circumvent this issue. What happens now is you're linking `processed_trait_datasets` with `trait_descriptors` on its primary key so you will only get those `trait_descriptors` which correspond to each `processed_trait_dataset`. This is not all of them - for example, with "erucic acid content" your query will only catch one (of three) `trait_descriptors` records.

[...]

PR #137, opened 2015-04-11
Linkage maps and QTLs data tables

Nuanda:
Adds two other tabs to the data tables view. The QTL query is "wrong" until Piotr performs the curation mentioned in #136 .

### DIFF
--- a/app/models/concerns/pluckable.rb
+++ b/app/models/concerns/pluckable.rb
@@ -1,30 +1,30 @@
-# Single Concern for all models to pluck certain columns for data tables
-# Pass columns in table like that:
+# Single Concern for all models to pluck columns for data tables
+# Define columns to be plucked by reloading:
+#
+#  - self.table_columns: columns visible to the end user
+#  - self.ref_columns: columns not visible to the end user, esp. references
+#
+# Columns should be defined in an array like that:
 # [
 #   'plant_line_name',
 #   'taxonomy_terms.name'
 # ]
 module Pluckable extend ActiveSupport::Concern
   included do
-    def self.pluck_columns(columns)
+    def self.pluck_columns
       query = self.all
-      columns = columns.map do |column|
-        if column.to_s.split('.').length == 3
-          from, to, field = column.to_s.split('.')
-          from = from.singularize unless reflections.keys.include?(relation)
-          to = to.singularize unless reflections.keys.include?(relation)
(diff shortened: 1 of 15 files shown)


Nuanda:
bump

app/models/concerns/pluckable.rb
```diff
@@ -16,8 +16,9 @@ def self.pluck_columns
       columns = table_columns + ref_columns
       columns.each do |column|
         relation = column.to_s.split('.')[0].pluralize if column.to_s.include? '.'
-        next unless relation and relation != self.table_name
+        next unless relation && relation != self.table_name
```
nowakowski: I use `and`. Is this a best practice?

app/models/linkage_map.rb
```diff
@@ -11,5 +11,24 @@ class LinkageMap < ActiveRecord::Base
...
 
+  default_scope { includes(plant_population: :taxonomy_term) }
+
+  include Pluckable
+
+  def self.table_data(params = nil)
+    pluck_columns
+  end
+
+  def self.table_columns
```
nowakowski: Out of curiosity - why these columns? Seems to be plenty more payload in `linkage_maps`.

app/views/data_tables/index.html.haml
```diff
@@ -11,5 +11,10 @@
       - table_column = model_param + '.' + table_column unless table_column.include? '.'
       %th= t("tables.#{table_column}")
 
+  - if model_class.respond_to? :count_columns
+    - model_class.count_columns.each do |count_column|
+      - count_column = count_column.scan(/\(([^\)]+)\)/).last.first
```
nowakowski: Bit puzzled. What is the intended behavior here?

app/models/concerns/pluckable.rb
Nuanda: Yes, most of the time && is a better choice (these are slightly different operators) but if you prefer and and it works, it's ok with me.

app/models/linkage_map.rb
Nuanda: These are from the original Interstore CS table.

app/views/data_tables/index.html.haml
Nuanda: To get only the content inside parentheses (). So from count(a.b) only a.b is taken, in order to deliver a proper join.

nowakowski:
:+1:

PR #138, opened 2015-04-15
Added model validations to match DB constraints

### LINKED ISSUE #20: Validations for constraints defined in DB schema are missing
Specifically - a lot of not null constraints is not reflected in validations. Just a reminder as you are probably aware of that.

nowakowski:
Fixes #20


app/models/genotype_matrix.rb
```diff
@@ -3,4 +3,16 @@ class GenotypeMatrix < ActiveRecord::Base
...
+
+  validates :original_file_name,
+            presence: true
+
+  validates :number_markers_in_matrix,
+            presence: true
+
+  validates :number_lines_in_matrix,
+            presence: true
 end
```
Nuanda: What about the `matrix` field. Looks like something important ;).

app/models/plant_accession.rb
```diff
@@ -5,4 +5,11 @@ class PlantAccession < ActiveRecord::Base
   has_many :plant_scoring_units
 
   include Annotable
+
+  validates :plant_accession,
+            presence: true
+
+  validates :year_produced,
+            presence: true,
+            length: {is: 4}
```
Nuanda: This one is actually _not_ required by the migration. It is ok with me - but did you check all current DB values in this column fit that constraint?

app/models/plant_population.rb
```diff
@@ -29,6 +29,13 @@ class PlantPopulation < ActiveRecord::Base
 
   include Filterable
 
+  validates :name,
+            presence: true,
+            allow_blank: true
+
+  validates :canonical_population_name,
+            presence: true
```
Nuanda: This one is not required according to the schema. If you think it should be required, then it's ok.

app/models/plant_trial.rb
```diff
@@ -8,6 +8,37 @@ class PlantTrial < ActiveRecord::Base
...
+
+  validates :project_descriptor,
+            presence: true
+
+  validates :plant_trial_description,
+            presence: true
+
+  validates :trial_year,
+            presence: true,
+            length: { is: 4 }
```
Nuanda: Again, the same comment as with the plant_accessions.year_produced.

app/models/plant_trial.rb
```diff
@@ -8,6 +8,37 @@ class PlantTrial < ActiveRecord::Base
...
+            presence: true
+
+  validates :trial_location_site_name,
+            presence: true
+
+  validates :place_name,
+            presence: true
+
+  validates :latitude,
+            presence: true
```
Nuanda: This one I would drop, together with the default and allow nil in the DB

app/models/plant_trial.rb
```diff
@@ -8,6 +8,37 @@ class PlantTrial < ActiveRecord::Base
...
+
+  validates :longitude,
+            presence: true
```
Nuanda: Same as for latitude.

app/models/primer.rb
```diff
@@ -5,9 +5,21 @@ class Primer < ActiveRecord::Base
   has_many :marker_assays_b, class_name: 'MarkerAssay',
              foreign_key: 'primer_b_id'
 
+  include Annotable
```
Nuanda: Why did you move this `include` here? There is an important reason why this is included in the last line of the model definition.

app/models/restriction_enzyme.rb
```diff
@@ -5,4 +5,12 @@ class RestrictionEnzyme < ActiveRecord::Base
...
              foreign_key: 'restriction_enzyme_b_id'
 
+  validates :restriction_enzyme,
+            presence: true
+
+  validates :recognition_site,
+            presence: true
+
+  validates :data_provenance,
+            presence: true
```
Nuanda: Please remove that and drop not null constraint from the DB.

Nuanda:
One general comment - please leave `include Annotable` as the last line in a model definition. As stated in the docs of that Concern, that is important.

app/models/primer.rb
nowakowski: Reason being? (I moved it to match the structure of all other model classes: (1) relationships; (2) interfaces; (3) validations; (4) custom methods.)

Nuanda: See my other comment about Annotable.

app/models/genotype_matrix.rb
nowakowski: Indeed; fixed.

app/models/plant_accession.rb
nowakowski: Yep, I did. They're all valid year designations.

app/models/plant_population.rb
nowakowski: All existing records have what looks like a proper value in this field, but you're right - it's not NN. I will remove this validation.

app/models/plant_trial.rb
nowakowski: All existing values are valid; I'm leaving this validation. I intend to submit another PR which converts datetime-like values to proper datetime types in the DB.

app/models/plant_trial.rb
nowakowski: Why, though? Existing records seem to have proper values in this field. Granted, it's not a GIS datatype, but Cropstore is not a GIS system so perhaps we don't need to roll out that particular cannon.

app/models/plant_trial.rb
nowakowski: See above.

app/models/primer.rb
nowakowski: Well, who knew. Fixing.

app/models/restriction_enzyme.rb
nowakowski: Rgr.

nowakowski:
All suggestions acknowledged, with the exception of comments concerning `latitude` and `longitude` in `plant_trials`. PR ready to be (re-)reviewed.

Nuanda:
:+1:

nowakowski:
Merged manually. Closing.

PR #139, opened 2015-04-15
Drops unneeded data_status column in linkage_groups

### LINKED ISSUE #51: Dropping data_status from entire database

nowakowski:
Fixes #51


Nuanda:
:+1:

PR #145, opened 2015-04-16
Removes obsolete table 'version'

### LINKED ISSUE #76: Do we need the `version` table?
The table seems to have tracked old changes in the CropstoreDB model. It doesn't seem to affect, or be in any way related to, the cropstore data model. I would suggest dropping it.

Alternatively we may retain the table and insert a record which reflects our current development work. If you prefer this option please tell us what the meaning of `08p` and `d5.10.4` in the `version` column is.

nowakowski:
Fixes #76


Nuanda:
:+1:

PR #148, opened 2015-04-16
Removes obsolete columns from countries table

### LINKED ISSUE #122: I don't think we need comments and data provenance in countries table

nowakowski:
Fixes #122


Nuanda:
:+1:

PR #150, opened 2015-04-16
Removes obsolete columns from trait_scores table

### LINKED ISSUE #123: Cleaning trait scores table
Trait scores table is of considerable size (100k+ records) and will probably grow fast. Therefore it would be beneficial to remove useless columns from it. I vote to remove the following:
 - replicate_score_reading: all values equal "1"
 - score_spread: value either empty or "none given"

nowakowski:
Fixes #123


Nuanda:
:+1:

PR #163, opened 2015-04-19
Move annotations to a configurable partial

### LINKED ISSUE #154: Construct 'Additional annotations' button and dialog to be reused in various submission elements
Do not show:
 - `entered_by_whom` (just set it directly to the current user email, when finalizing the submission)
 - `date_entered` (just set it directly to the current date, when finalizing the submission)

Use the new dialog in PP submission:
 - in the last step
 - in the new PL form.

kammerer:
Fixes #154 

It is not 100% ready yet but the only thing missing is to decide which annotation fields need to be shown. You are welcome to comment though.


Nuanda:
The following annotation fields should be shown:
 - data_owned_by (but be aware this one might not be present for a given model)
 - data_provenance (always present for Annotables)
 - comments (always present for Annotables)

in this exact order.

Also remember the other two fields, entered_by_whom and date_entered, should _always_ be set (with the current user email and the current date), no matter if the user adds any other annotation or not.

app/views/submissions/_annotations.html.haml
```diff
@@ -0,0 +1,22 @@
+- unless defined?(form)
+  - form = nil
```
Nuanda: Can't we use: `- form ||= nil` ?

kammerer:
I used a collapsible element not a dialog - I feel it is more appropriate here but I will change it if you do not agree.

Nuanda:
If that works ok for embedded sub-forms, like adding a new PL, then it's ok with me.

Nuanda:
:+1:

PR #164, opened 2015-04-20
Removes pseudo-null values from model tables

### LINKED ISSUE #133: Implicit NULLs in certain model fields
Some records appear to "smuggle" blank values in NN (non-nullable) fields - this is evidenced by the presence of `n/a` or `na` string literals.

Suspect models include:

- `qtl.map_qtl_label` (52 records listing `n/a`)
- `design_factors.design_factor_2` (7 records listing `na`)
- `design_factors.design_factor_3` (7 records listing `na`)
- `design_factors.design_factor_4` (2431 records listing `na`)
- `design_factors.design_factor_5` (2431 records listing `na`)

My suggestion would be to drop the NN restriction in the affected fields and convert the above mentioned values to NULL.

nowakowski:
Fixes #133


Nuanda:
:+1:

Issue #161, opened 2015-04-19
Proper redirect for non-logged "Submit new X" button click


Nuanda:
Currently the system redirects to the main page, which seems counter-intuitive. It should stay on the submission page.

Nuanda:
A new problem: `Cannot redirect to nil!` on session timeout.

Issue #165, opened 2015-04-20
Footer height problem


Nuanda:
I adjusted `@include sticky-footer(10px, "#wrap");` call in the `app/assets/stylesheets/partials/_footer.scss` file to temporary alleviate the problem (it was not possible to submit user data due to the footer taking too much space) but you probably want to look at the issue anyway and fix it the way it should be.

PR #167, opened 2015-04-21
Sass styles

bchucherko:
Footer problem should fix issues #161 #165


Nuanda:
I don't think it does fix #161 but then again I don't think it should - you probably meant some other issue. It does fix the footer height issue, though.

Nuanda:
:+1:

PR #170, opened 2015-04-21
Search action

### LINKED ISSUE #140: Setup search action for the front page search bar and present found results
For now we may list found plant_populations, plant_lines and plant_varieties. I need information what exactly do you require on the datatables side. I.e. when 7 plant lines are found, and the user clicks on that result, s/he should be able to land on the plant lines data tables view.

I need to know how the filter parameters for that exact set of plant lines are passed to the controller.

kammerer:
Fixes #140 

After selecting search result item labelled with "Found N models" user is redirected to data table view which additionally gets search term as a param, e.g. http://lvh.me:3000/data_tables?model=plant_varieties&search=plepleple

We should probably replace the part of `Filterable` module's which is responsible for handling `params[:search]` with ES code.


Nuanda:
Tomek, I think I might have miscommunicated the way the search bar should be implemented. It is not meant to work as an autocomplete selection. When you go to the UXPIN wireframes prepared by @bchucherko you will find several examples there.

When you look at the basic v2 version you'll see, that it is a plain search bar (perhaps it is a good idea to limit search to words that are e.g. at least 3 letters long). The results are presented in a drop down fashion just below.

Regarding the examples, I think we should go with the v4 version, but that is for later - let's implement the basic search functionality now.

In the meantime I'll try to implement data tables support for showing chosen search results in a table.

kammerer:
Oops, my bad. Fortunately this should be doable with minor changes in code.

May come in handy:
https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-model#search-results-as-database-records
I think that reusing `Search` is important since the exact ES query we use
may be changed yet (e.g. to boost importance of some fields).

2015-04-22 9:49 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Tomek, I think I might have miscommunicated the way the search bar should
> be implemented. It is not meant to work as an autocomplete selection. When
> you go to the UXPIN wireframes prepared by @bchucherko
> <https://github.com/bchucherko> you will find several examples there.
>
> When you look at the basic v2 version you'll see, that it is a plain
> search bar (perhaps it is a good idea to limit search to words that are
> e.g. at least 3 letters long). The results are presented in a drop down
> fashion just below.
>
> Regarding the examples, I think we should go with the v4 version, but that
> is for later - let's implement the basic search functionality now.
>
> In the meantime I'll try to implement data tables support for showing
> chosen search results in a table.
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/170#issuecomment-95064903>.
>



-- 
Tomasz Szymczyszyn

Nuanda:
Sure I was going to reuse the new Search service class in my controller/model - and I guess we need to assume the field priority/importance does not change between the initial search call and the datatables display. Or am I missing something important in the link you have posted?

kammerer:
The part I thought may be useful in the link was the ability to chain AR
calls to ES search results.

2015-04-22 10:20 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Sure I was going to reuse the new Search service class in my
> controller/model - and I guess we need to assume the field
> priority/importance does not change between the initial search call and the
> datatables display. Or am I missing something important in the link you
> have posted?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/170#issuecomment-95071015>.
>



-- 
Tomasz Szymczyszyn

Nuanda:
Won't we break submissions' autocomplete fields when we replace Filterable's search param handling with ES code?

Nuanda:
I also found alternative code in `app/forms/concerns/model_helper.rb` to search for PLs. Do you use that one for submissions' autocomplete now, instead of Filterable?

kammerer:
It should not break but may give somewhat different results.

Autocomplete does not use `ModelHelper`. It was only used to populate
initial values in submission views. I will review that, maybe it should be
removed to avoid confusion.

2015-04-22 10:37 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Won't we break submissions' autocomplete fields when we replace
> Filterable's search param handling with ES code?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/170#issuecomment-95077614>.
>



-- 
Tomasz Szymczyszyn

Nuanda:
Is it ok if I commit to this branch my changes regarding handling search result display in datatables?

kammerer:
Sure

2015-04-22 11:23 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Is it ok if I commit to this branch my changes regarding handling search
> result display in datatables?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/170#issuecomment-95089675>.
>



-- 
Tomasz Szymczyszyn

Nuanda:
The main problem I have in mind is that we ES-search with multiple columns. For instance, in the case of PLs, we also ES-search by previous_plant_line names. For autocomplete this may result in wrong suggestions... So I'm still not 100% sure about reusing that part of Filterable.

kammerer:
You are right, it would be very strange to see in autocomplete sth that
does not match entered string. Do you think it will not affect data table
views? I mean, will all filterable fields be shown there?

2015-04-22 12:00 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> The main problem I have in mind is that we ES-search with multiple
> columns. For instance, in the case of PLs, we also ES-search by
> previous_plant_line names. For autocomplete this may result in wrong
> suggestions... So I'm still not 100% sure about reusing that part of
> Filterable.
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/170#issuecomment-95109221>.
>



-- 
Tomasz Szymczyszyn

Nuanda:
That's another thing to consider: #172. However that will not be a problem.

For the time being I'll change ES-search param name to "fetch" (only for DataTables controller call) and  will support that param in Filterable.filter.

kammerer:
Sounds good.

2015-04-22 12:58 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> That's another thing to consider: #172
> <https://github.com/eSpectrum-IT/brassica/issues/172>. However that will
> not be a problem.
>
> For the time being I'll change ES-search param name to "fetch" (only for
> DataTables controller call) and will support that param in
> Filterable.filter.
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/170#issuecomment-95137515>.
>



-- 
Tomasz Szymczyszyn

Nuanda:
BTW PL.as_indexed_json is public but PP.as_indexed_json is private. Please make it consistent for all models, one way or another.

Nuanda:
I implemented the model/Filterable part. One notable thing is that I am not able to test the real ES-based fetch - see the last test in `models/concerns/filterable_spec.rb`. Nor do I know if I need to - because the Search service might be tested elsewhere later. So I leave it up to you - either remove that test altogether or use it accordingly.

Nuanda:
Well, works much nicer now. Still there is the question of performing autocomplete-like search (as it is now) or wait for the user to press the Search button.

I opted for the later - for I fear that after we indexed entire DB, the autocomplete style would be too demanding for the server (performing a lot of searches while the user types). But we may also leave it as it is, in order to (1) show it to the user and ask what they think and (2) perform the entire DB indexing and testing it against all data.

Nuanda:
Also, something is wrong, at least with my setup. When searching for `napus` I get only a single PP (this one has `napus` in its description attribute). According to the model ES settings I should also get all PPs that are related to TaxonomyTerm with name containing `napus`. I did app:bootstrap just prior to that.

Should I, perhaps, somehow _reset_ ES index?

kammerer:
Sorry, I do not know why I made it behave like autocomplete in the first place. Updated yet again. It still uses XHR but requires user to submit query by pressing enter / clicking button.

You can run `bin/rake environment elasticsearch:import:model CLASS=PlantPopulation FORCE=y` to reset index.

kammerer:
One more thing, I changed the ES query to include wildcards at both ends of given term. Before it only looked for words beginning with that term.

kammerer:
Okie dokie, finally found it - missing PPs were caused by making `#as_indexed_json` private :/ I changed spec setup so that `Search` could finally be tested and started implementing its specs. I will add a separate task reminding to complete its pending specs.

Nuanda:
:+1:

PR #175, opened 2015-04-22
Fixes population_loci records with incorrect plant_population values

### LINKED ISSUE #105: Unknown plant population names in population_loci
There are two `population_loci` with population names which do not exist in `plant_populations`

  - BolNGDH_02 (there exists a `plant_populations` record named `BolNGDH_01`)
  - BraVCS3M_0 (there exists a `plant_populations` record named `BraVCS3M_01`)

Could you suggest what to do with these records?

nowakowski:
Fixes #105

### DIFF
--- a/app/models/population_locus.rb
+++ b/app/models/population_locus.rb
@@ -6,6 +6,9 @@ class PopulationLocus < ActiveRecord::Base
   has_many :map_positions
   has_many :map_locus_hits
 
+  validates :plant_population,
+            presence: true
+
   validates :mapping_locus,
             presence: true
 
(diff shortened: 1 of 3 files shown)


app/models/population_locus.rb
```diff
@@ -6,9 +6,6 @@ class PopulationLocus < ActiveRecord::Base
   has_many :map_positions
   has_many :map_locus_hits
 
-  validates :plant_population,
-            presence: true
```
Nuanda: Shouldn't that stay? The relation is still present, right?

nowakowski: OK; I'll keep it.

Nuanda:
:+1:

PR #177, opened 2015-04-22
Submission submitted object binding

### LINKED ISSUE #109: PPSubmissionDecorator problem with no relation to submitted PP
Currently when PPsubmission is completed, it is not assigned to the created PP object. It prevents PPSubmissionDecorator.details_path from getting link to the submitted PP object.

Assuming every submission will submit one "main" (or "parent") object, perhaps we may solve it by having a single "generic" relation id to a submitted object? Still, it is not clear at the moment if this is a good solution, so for the time being we need to see what happens in further submission processes.

Nuanda:
Fixes #109 

Fix is generic, should work for future submission types as well.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -65,16 +65,16 @@ class Submission < ActiveRecord::Base
 
   def associated_model
     case submission_type
-      when 'population'
-        PlantPopulation
-      when 'traits'
-        PlantScoringUnit
-      when 'qtl'
-        Qtl
-      when 'linkage_map'
-        LinkageMap
-      else
-        nil
+    when 'population'
+      PlantPopulation
+    when 'traits'
+      PlantScoringUnit
+    when 'qtl'
+      Qtl
+    when 'linkage_map'
+      LinkageMap
+    else
+      nil
     end
   end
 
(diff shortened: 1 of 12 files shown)


app/models/submission.rb
```diff
@@ -58,6 +63,21 @@ def name
     [I18n.l(created_at, format: :short), content.step01.try(:name)].compact.join(' ')
   end
 
+  def associated_model
+    case submission_type
+      when 'population'
```
kammerer: Strange indentation ;)

spec/factories/submission.rb
```diff
@@ -2,5 +2,12 @@
   factory :submission do
     submission_type :population
     user
+
+    trait :finalized do
```
kammerer: I introduced such trait too in #178 but it is more havy-weight  as it assigns `submission.content` and performs finalization. I think it is worth to keep it but maybe under different name.

kammerer:
:+1:

app/models/submission.rb
Nuanda: Fixed.

spec/factories/submission.rb
Nuanda: Let's do it like that: I merge my PR and when merging yours, you will add what is important (mainly: setting submitted_object_id) from my trait to your (better) trait, and after all specs are ok, you'll remove my trait.

Issue #114, opened 2015-04-03
Unknown scoring occasion names in trait_scores


nowakowski:
There are fifteen distinct scoring occasion names in `trait_scores` which do not correspond to any record in `scoring_occasions`:

  - occ_HRI_UN0000_003
  - occ.WHRI_MB0000_002
  - occ.RRES_JB0410_002
  - occ.RRES_JB0310_001
  - occ.WHRI_MB0000_001
  - occ.RRES_JB0410_001
  - occ_PRT7_UN0000_001
  - occ.WHRI_MB0000_003
  - occ_PRT7_UN0000_002
  - occ_HRI_UN0000_001
  - occ.WHRI_MB0000_004
  - occ_HRI_UN0000_002
  - occ_HRI_UN0000_004
  - occ.RRES_JB0310_002
  - occ_PRT7_UN0000_003

I have linked `trait_scores` records to `scoring_occasions` whenever a match can be found (78083 records) and left the FK field null in other cases (31142 records). I have also left the `scoring_occasion_name` field in `trait_scores` to enable future curation. Please tell us if you agree with this approach.

wjurkowski:
Sounds perfect, thumb up!

2015-04-03 13:20 GMT+01:00 Piotr Nowakowski <notifications@github.com>:

nowakowski:
Happy to hear. :)

Nuanda:
See #155 .

nowakowski:
Fixed in #181

PR #181, opened 2015-04-23
Drop scoring_occasions

### LINKED ISSUE #155: Replace scoring_occasion table with one new column inside the trait_scores table
Essentially, the entire scoring_occasion table holds four dates - all else is probably noise (i.e. useful information repeated in other tables, like institute or provenance).

I propose to drop that table and introduce a singe (optional) scoring_date field to trait_scores table, and populate it with score_start/end_date (both are always the same) from the dropped table.

nowakowski:
Fixes #155

### DIFF
--- a/app/models/scoring_occasion.rb
+++ b/app/models/scoring_occasion.rb
@@ -1,9 +1,0 @@
-class ScoringOccasion < ActiveRecord::Base
-
-  has_many :trait_scores
-
-  validates :scoring_occasion_name,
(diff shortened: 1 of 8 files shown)


nowakowski:
Rescues score_start_date which it writes to trait_scores as a new column (scoring_date).

nowakowski:
Also fixes #114

Nuanda:
:+1:

PR #182, opened 2015-04-23
Search vs tables setup cleanup

### LINKED ISSUE #173: ES model settings adjustments
Please:
 - [x] make as_indexed_json method visibility scope consistent (either private or public) in all searchable methods
 - [ ] remove :id fields as searchable since @nowakowski turned PKeys into proper DB-incremented integers and it does not make any sense any longer for the user to search these.

Nuanda:
Fixes some issues with the current ES setup in PlantPopulation.

Makes sure all ES-searchable fields are displayed in DataTables so a user is not confused why a given record was found in response to her query.

Fixes #173


kammerer:
:+1:

PR #186, opened 2015-04-24
Generic related column in datatables

### LINKED ISSUE #183: Design generic solution to navigate from one DT to related DT by number of related models

Nuanda:
This one's quite complex - I hope the added complexity is worth it and will make designing further data tables for the rest of CS model faster and nicer.

Relatable concept redefines model's count_columns the was, that it provides a dropdown to quickly navigate to other datatables via relations. So when e.g. PP has_3 PLs, the DT dropdown button will include "3 plant lines" link, which should relate user to PL datatable with the 3 PL records filtered.

It relies on cached rails counters. It is relatively complex on the coffeescript side, since the dropdown button creation takes place entirely in js ajax response handling (through a Datatable mechanism).

Please note I decomposed already long data_tables.coffee into 3 scripts: DT's initializers, column handlers applicable to many models, and column handlers applicable to specific models.

Fixes #183


spec/models/concerns/relatable_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    pp = create(:plant_population,
+                linkage_maps: create_list(:linkage_map, 3, plant_population: nil))
+    expect(pp.linkage_maps.count).to eq 3
+    expect(pp.reload.linkage_maps_count).to eq 3
+    pp.linkage_maps << create(:linkage_map, plant_population: nil)
+    expect(pp.linkage_maps.count).to eq 4
+    expect(pp.reload.linkage_maps_count).to eq 4
+    lms = create_list(:linkage_map, 2, plant_population: nil)
+    pp.linkage_maps = lms
+    expect(pp.reload.linkage_maps.count).to eq 2
```
kammerer: Unintended `reload`?

Nuanda: Intentional. This is how counter cache works - it does NOT work until you draw the model from the DB again.

kammerer:
:+1: Though I wonder if we should not be using ES facets for such stuff.

Nuanda:
I don't know. This is quite "natural" and "straightforward" functionality. It's just the generic approach we need (due to multitude of tables) that makes it kind of complex.

spec/models/concerns/relatable_spec.rb
kammerer: Oh, I overlooked the difference between line 21 and 22.

PR #195, opened 2015-04-26
Gets rid of old default values in annotation columns

### LINKED ISSUE #189: Remove from DB current default values for 'annotations' columns
See #125 for the list of "annotation" fields (that repeat a lot for various tables in CS) and the list of tables that have them. Note, however, that recent data curations might have shrunk that table list.

I think the problem mainly regards the notorious 'No comment' "value" in the comments column.

nowakowski:
Fixes #189


Nuanda:
:+1:

PR #201, opened 2015-04-29
Extends TraitDescriptor.table_data to count QTLs for each TD

### LINKED ISSUE #199: Extend TraitDescriptor.table_data query to count associated QTLs
The counting has to go through the PTDs. Please add the qtls count as the penultimate result column.

nowakowski:
Fixes #199


nowakowski:
**Caution:** Since `TraitDescriptor.table_data` enumerates plant_trials rather than trait_descriptors, the same counter value will be listed for multiple records (because each trait descriptor may appear in multiple rows). Also note that there are some `trait_descriptors` which are not bound to any `plant_trials` and therefore will not appear in the table - and neither will their associated QTLs.

PR #202, opened 2015-04-30
New data tables

### LINKED ISSUE #199: Extend TraitDescriptor.table_data query to count associated QTLs

Nuanda:
A lot of things inside. Most important are:
 - new, more functional QTL table
 - brand new PlantAccessions and QTLJobs tables
 - changed TD table
 - some new relations between these and other tables
 - refactoring here and there.

@kammerer Please see the changes I made to spec/services/search_spec - I don't know precisely why the db cleaning did not work in my case, but I was getting a lot of TT.name uniqueness validation violation in specs.

Fixes #199


Nuanda:
Please, *DO NOT* delete this particular branch after merging with master, I intend to use it further.

kammerer:
No idea why `DatabaseCleaner` does not work for you. Perhaps test DB was not empty when you run specs (but I don't know how would you end up with these particular TTs in the DB)? Can you try adding these at the top of `before(:all)` block?
  
  DatabaseCleaner.strategy = :truncation
  DatabaseCleaner.clean

Just to check if this works as it is probably better to explicitly delete created records as you did  instead of truncating all tables.

app/models/concerns/relatable.rb
```diff
@@ -28,6 +28,8 @@ def self.get_related_model(counter_name)
       counter_name = counter_name.split(/ as /i)[-1]  # honor aliasing
       counter_name.gsub!('_count','')   # support cached count columns as well
 
+      counter_name.gsub!('qtls','qtl')  # unfortunate special case... :(
```
kammerer: Shouldn't we rename `qtl` table to `qtls`?

app/models/linkage_group.rb
```diff
@@ -20,5 +19,41 @@ class LinkageGroup < ActiveRecord::Base
...
+
+  def self.count_columns
+    [
+      'linkage_groups.map_linkage_group_lists_count AS linkage_maps_count'
+    ]
+  end
+
+  private
+
+  def self.permitted_params
```
kammerer: That's what I was talking a few days ago - it does not matter if you put `self.permitted_params` below `private` or not. It is always public.

app/models/trait_descriptor.rb
```diff
@@ -21,8 +21,12 @@ class TraitDescriptor < ActiveRecord::Base
             numericality: true
 
   def self.table_data(params = nil)
+    trait_descriptor_query = ''
+    if params && params[:query].present? && params[:query][:id].present?
+      trait_descriptor_query = "WHERE td.id = #{params[:query][:id]}"
```
kammerer: SQL injection vulnerability?

spec/models/plant_accession_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      pas = create_list(:plant_accession, 2)
+      filtered = PlantAccession.filter(
+        query: { id: pas[0].id }
+      )
+      expect(filtered.count).to eq 1
+      expect(filtered.first).to eq pas[0]
+    end
+  end
+
+  describe '#pluckable' do
```
kammerer: #pluck_columns

spec/models/plant_scoring_unit_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'rails_helper'
+
+RSpec.describe PlantScoringUnit do
+  describe '#pluckable' do
```
kammerer: #pluck_columns

kammerer:
QTL datatable does not fit in on my display, but I do not know how to solve this.
One fairly important issue (SQL injection) + some minor comments, otherwise :+1:

app/models/linkage_group.rb
Nuanda: Should I remove the `private` keyword from every model which does not define private object methods?

app/models/trait_descriptor.rb
Nuanda: Yes, you're probably right. Added simple .to_i - I hope this is enough.

spec/models/plant_accession_spec.rb
Nuanda: Fixed here and in all other specs.

app/models/concerns/relatable.rb
Nuanda: Well, that would certainly help with some things, but I guess it is rather late for that kind of changes. @nowakowski decided to do it that way (for some reasons) so I tried to adhere to that.

nowakowski: The "l" in `qtl` stands for "locus", and the plural of "locus" is "loci", so `qtls` is just bad grammar.

nowakowski: ...besides, the table was called `qtl` to begin with and I didn't really see a reason to change it. I suppose it could be changed to `quantitative_trait_loci` but then again, I've recently been receiving complaints about coming up with overly winding names. :)

Nuanda:
Well, seems it was a kind of transient problem with the search_spec - reverted my changes since all looks to be all right again there.

All other things fixed according to comments (besides the qtl - qtls issue). The QTL table width issue is known, and noted down in #213 for further work.

Please check my fixes.

app/models/concerns/relatable.rb
kammerer: I am not going to force it, especially that it is late indeed, but I do not think that these arguments are valid. First, we changed a lot in the database schema so why not this particular name? The reason for this would be to go with the convention which says table names are plural. As for grammar, I think it is rather common practice to make plural forms of acronyms. Also, [wikipedia says it is correct](http://en.wikipedia.org/wiki/Quantitative_trait_locus) ;)

kammerer:
Looks good :+1:

app/models/concerns/relatable.rb
nowakowski: @kammerer "We changed a lot, so why not this?" is a non sequitur. The fact that we introduce a lot of changes is not an argument in favor of any particular change. Frankly, I was unaware that this was such a big headache for the upper layers of the stack. I am happy to rename the table even now, if you so desire.

You are right in that acronyms are sometimes pluralized in the way cited.

kammerer: @nowakowski Of course that was not the reason, it was in the next sentence ;P I wouldn't myself know upfront how that affects other layers. I think it could be added as a low priority issue, WDYT?

nowakowski: @kammerer Added #215. Will fix, time permitting. :)

kammerer: Great :)

2015-05-02 21:52 GMT+02:00 Piotr Nowakowski <notifications@github.com>:

> In app/models/concerns/relatable.rb
> <https://github.com/eSpectrum-IT/brassica/pull/202#discussion_r29550166>:
>
> > @@ -28,6 +28,8 @@ def self.get_related_model(counter_name)
> >        counter_name = counter_name.split(/ as /i)[-1]  # honor aliasing
> >        counter_name.gsub!('_count','')   # support cached count columns as well
> >
> > +      counter_name.gsub!('qtls','qtl')  # unfortunate special case... :(
>
> @kammerer <https://github.com/kammerer> Added #215
> <https://github.com/eSpectrum-IT/brassica/issues/215>. Will fix, time
> permitting. :)
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/202/files#r29550166>.
>



-- 
Tomasz Szymczyszyn

Nuanda: Guys, thanks for having this long, academic, rhetoric, opinionated, colorful, firebrand discussion about how to add me a lot of work. I would much prefer to have that kind of discussions taking place, like, 03.03, instead of two months later :).

Still, if Piotr wants to change table, model and has_many relations names from qtl to qtls, please be it rather sooner than later!

PR #210, opened 2015-05-02
API

### LINKED ISSUE #197: Implement rudimentary API
For starters:
* API key authentication
* a few sample controllers

kammerer:
Read API:

* more or less ready for `PlantPopulation`, `PlantLine` and `PlantVariety`
* allows to fetch collection or single object
* exposing a model is as simple as adding it in API initializer (assuming it already is `Filterable`)   
* users authenticate with API key (key is created automatically for each new user)
* pagination

Fixes #197


Nuanda:
@kammerer Tomasz, please:
 - do not tackle create / update API in this PR, it would be too much
 - ref_columns are *not* needed in API, they are needed for tables to be able to construct proper URLs
 - however it is *important* to include object's `id` in the returned attributes, since the client would probably need it for further API calls
 - actually, I would not constrain get API returns with table_columns - can't we simply bundle *all* attributes in the returned json?
 - when creating API Key view, do not forget to provide `regenerate` button, so a user may revoke a compromised key (the first key might be generated on user object creation)
 - no need for `except: :new` in API controller since there is probably no :new action for API, right?

Side loading for has_many / belongs_to is up to you. However, it would be even better to get a first version of rudimentary API Docs view - can we do it that way? I.e.: simplest read queries, API key management view - then switch to API docs right away - and then going back for side loading, write API, sandbox?

Having a simple API Docs view makes it ready for Bartosz to style it, while we can add API docs subsections and change text content later.

kammerer:
I will then create new issues for missing functionality and make this PR ready for review.

I have already added some API documentation in [api-docs](https://github.com/eSpectrum-IT/brassica/tree/api-docs) branch. Let me know what do you think about its format.

Nuanda:
All right, lets do that.

kammerer:
@Nuanda Fixed some of the issues you mentioned.

app/controllers/api/v1/resources_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  def model_name
+    @model_name ||= request.path.match(/\A\/api\/v1\/(([\w_]+)\/?)/)[2]
+  end
+
+  def model_klass
+    @model_klass ||= model_name.singularize.camelize.constantize
+  end
+
+  def allowed_models
+    Brassica::Api.models.map { |klass| klass.name.underscore.pluralize }
```
Nuanda: I think klass.table_name should do the same. It will also take care for cases like qtl/qtls (while pluralize might get it wrong).

app/controllers/api/v1/resources_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      raise ActionController::RoutingError.new('Not Found')
+    end
+  end
+
+  def model_name
+    @model_name ||= request.path.match(/\A\/api\/v1\/(([\w_]+)\/?)/)[2]
+  end
+
+  def model_klass
+    @model_klass ||= model_name.singularize.camelize.constantize
```
Nuanda: I'm not 100% sure but I think classify is a shortcut for singularize.camelize.

config/initializers/api.rb
```diff
@@ -0,0 +1,16 @@
+module Brassica
+  module Api
+
+    def self.models
+      (self.readable_models + self.writable_models).uniq
```
Nuanda: I think `self.readable_models | self.writable_models` will do the same.

app/controllers/api/v1/resources_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  before_filter :authenticate_api_key!
+  before_filter :require_allowed_model
+
+  # FIXME find a way to document api calls
+
+  def index
+    filter_params = params[model_name.singularize].presence
+
+    resources = filter_params ? model_klass.filter(filter_params) : model_klass.all
```
Nuanda: Some models do not include Filterable. But perhaps it is enough to have a spec that tests that all API.readable_models are Filterable as well.

Nuanda:
Very minor remarks - consider them at your discretion. Merge afterwards. :+1:

app/controllers/api/v1/resources_controller.rb
kammerer: I think there are API specs which would fail for these classes but I am adding a spec to make that requirement explicit.

app/controllers/api/v1/resources_controller.rb
kammerer: Table name does not matter here, this string is compared to the part of url which describes model. For API consistency I would make it '/api/v1/qtls'.

PR #211, opened 2015-05-02
Sass styles

bchucherko:
Pretty much everything is done according to email.
Some remarks:
1. Popovers now include proper styling, however please check under FF as they haven't been working in this browser
2. Buttons like 'back to plant lines', 'see all' should IMO be placed near the table lead, not withing the table-bar. This not only might be a bit confusing, but also poses a problem when multiple buttons are present and there's not enough space. I don't now however how to move them there.
3. I haven't done anything with table width, but this will be an issue later (maybe we will have to use .table-responsive) 
4. I fixed some errors that I found along the way
5. Added 'inview' script so that piechart will animate when user scrolls to it.


Nuanda:
Thanks a lot. Noted table bar buttons suggestion in #212. Tested popover in FF and seems to work ok.

Nuanda:
:+1:

PR #221, opened 2015-05-03
Basic read API documentation structure

### LINKED ISSUE #220: Create basic API documentation structure

kammerer:
Needs drying up and working out details but comments are welcome.

I imagine the outline as follows:
* general introduction (authentication, format, pagination, conventions, etc)
* specific resources
  * resource A
    * description
    * attributes
    * action X
      * url
      * request params
      * example request
      * button to call the API and show response (for users signed in)
      * example response (for users not signed in - not added yet)
    * action Y
    * [...]
  * resource B
  * [...]

Fixes #220


app/models/plant_population.rb
```diff
@@ -87,9 +87,7 @@ def as_indexed_json(options = {})
   def self.permitted_params
     [
       :fetch,
-      query: [
-        :id, :name
-      ]
+      query: [:id, :name]
```
Nuanda: In my PR I decided, mainly for clarity and consistency, to make all query parameters strings, not symbols (some include a dot, and keeping them all symbols would require ugly syntax..).

Will switching that to strings, not symbols, be of any problem to this PR?

Nuanda:
Tomasz, keeping in mind our wireframes for that: https://live.uxpin.com/3d91dec11a9043767b221f04117dcdfd221c864a#/pages/19058668, I suggest we should stick to single-level section hierarchy. So let's have:

 - general introduction (...): the thing the user sees instantly when navigating to that part
 - fetching data: still generic (i.e. cross-model) section about how GET queries look like, how to pass arguments, what to expect as results
 - searching for data: similar generic section about how to issue full-text search calls to BIP, how to adjust scope (model), what to expect as the outcome
 - resource A (attributes, allowed query parameters, full-test searching, examples)
 - resource B (...)

Having it in such simple structure will help Bartosz to put it into a proper styling.

BTW - can we also support entire BIP text search through API (same operation that is used in the front page search panel)?

app/models/plant_population.rb
kammerer: I will revert that so we don't get conflicts. Anyway, it's only formatting change.

kammerer:
@Nuanda What you suggest is more or less what I had in mind. Nesting in my outline does not indicate navigation links or separate subpages.

kammerer:
While the content itself is still very sketchy I think that there are all important elements included.

kammerer:
Oops, I've broken the PR by rebasing my branch. Don't know how to fix it so I will make a new one...

Issue #227, opened 2015-05-05
MapLocusHit unnecessary data status columns cleanup


Nuanda:
Piotr, MLH table has three columns:
 - `cs_sequence_data_status`
 - `sqs_sequence_data_status`
 - `map_data_status`

and all of their values read either `public` or `published`. Since we have already did that with the `data_status` column in other tables, I think we may also safely drop these three columns here as well.

PR #230, opened 2015-05-06
Removed redundant columns from map_locus_hits

nowakowski:
Solves #227


Nuanda:
Have you checked existing model validations for that columns?

nowakowski:
Sorry, I was distracted. Good catch.

Nuanda:
:+1:

PR #234, opened 2015-05-06
final datatables for MP, PLoci, MLH and their relations

### LINKED ISSUE #224: Add three more yellow datatables after all related data curation issues are solved

Nuanda:
Fixes #224

### DIFF
--- a/spec/models/plant_accession_spec.rb
+++ b/spec/models/plant_accession_spec.rb
@@ -5,7 +5,7 @@ RSpec.describe PlantAccession do
     it 'will query by permitted params' do
       pas = create_list(:plant_accession, 2)
       filtered = PlantAccession.filter(
-        query: { id: pas[0].id }
+        query: { 'id' => pas[0].id }
       )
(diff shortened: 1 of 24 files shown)


Nuanda:
@nowakowski Piotr, I merged your latest data fixes to this PR.

spec/models/plant_accession_spec.rb
```diff
@@ -5,7 +5,7 @@
     it 'will query by permitted params' do
       pas = create_list(:plant_accession, 2)
       filtered = PlantAccession.filter(
-        query: { id: pas[0].id }
+        query: { 'id' => pas[0].id }
```
nowakowski: Why the old-fashioned notation?

Nuanda: I decided to turn all query parameters to string - that was discussed with Tomek as well. It makes things a little bit cleaner. With strings, however, you can't use the new notations (which is suitable for symbols).

nowakowski:
:+1:

Issue #232, opened 2015-05-06
LinkageMap has two redundant FKEY declarations?


Nuanda:
has_many :genotype_matrices, foreign_key: 'linkage_map_id'
  has_many :map_locus_hits, foreign_key: 'linkage_map_id'

Are foreign_key declarations here necessary?

PR #235, opened 2015-05-06
Miscellaneous curation fixes

### LINKED ISSUE #233: Problem with lack of proper MLH table primary key(?)
I think this is related to the lack of PKEY in that table, but perhaps the cause is different. Anyway, doing a simple operation like that:

`LinkageGroup.first.map_locus_hits << MapLocusHit.first`

results in:

`TypeError: nil is not a symbol nor a string`

exception. Can you please try to find out what is happening there?

nowakowski:
Fixes #233

### DIFF
--- a/app/models/linkage_map.rb
+++ b/app/models/linkage_map.rb
@@ -6,9 +6,9 @@ class LinkageMap < ActiveRecord::Base
 
   has_many :map_linkage_group_lists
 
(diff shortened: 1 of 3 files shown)


nowakowski:
Also fixes #232

Nuanda:
:+1:

PR #240, opened 2015-05-07
Sass styles

bchucherko:
1) step wizard styled
2) select2 library unified to suit bootstrap theme
3) overall improvements


Nuanda:
A couple of requests:
 - please set the New plant line form required labels with the same asterisk indicator as for other fields
 - I think it is safe now to remove the "Optional" hints below form controls
 - the `My submissions` view title should have proper style (I mean the green underline).

kammerer:
What do you think about changing "Add new Plant Line" subform to a two-column layout?

Nuanda:
In case you ask me, I have no strong opinion on that. If @bchucherko thinks this is doable and would look better, please do go ahead.

bchucherko:
Fixes & other pages styling

bchucherko:
As it comes to dividing the column - I see no point in doing that - because we have input...'or'...input?Could you elaborate?
Then why not do the same for the rest of forms?
Two column layout will have to fold eventually on smaller screens.

Nuanda:
@kammerer are you able to merge what @bchucherko did to the API docs view with your new version of that view, merged already to master?

bchucherko:
Sth went wrong while merging branches. I'll push another change soon

bchucherko:
Ready to merge.

kammerer:
WRT two-column layout, select or text input is one reason, the other is that form fields which are so wide look slightly weird to me (talking about medium-large display). Just my opinion, I am not saying this should be done and certainly should not block this PR.

kammerer:
I will take care of API docs.

kammerer:
I brought back API docs contents and added some missing styling.
@bchucherko Please, feel free to tweak as you like. Also, I did not manage to make scroll-spy for the side navigation work, can you help?

Nuanda:
I need this PR in master for deployment and testing. So I merge it right now and will make an issue for @bchucherko regarding the scroll spy.

PR #244, opened 2015-05-08
API key management view

### LINKED ISSUE #217: API key management view
User should be able to create a new API key.

kammerer:
Fixes #217 

Note: I wanted to use 'renew' as action name but it caused really strange issues in specs :/


app/views/application/api.html.haml
```diff
@@ -21,7 +21,7 @@
         BIP provides HTTP REST-style API which uses JSON to represent returned
         resources. Full access to the API is allowed for all registered users.
         In order to authenticate users pass <code>api_key</code> parameter in
-        each call to the API.
+        each call to the API. Sign in and get your key <a href='/api_keys'>here</a>.
```
Nuanda: Can you use the link_to and a path helper here?

app/views/api_keys/show.html.haml
```diff
@@ -0,0 +1,14 @@
...
+
+  %p Your can access BIP API using the following key:
+
+  .row
+    .col-sm-6
+      %code
+        = @api_key.token
+
+    .col-sm-6
+      = button_to "Renew", recreate_api_keys_path, method: "put", class: 'btn btn-default'
```
Nuanda: Please add the following warning text:

"This operation will revoke your current API Key. That means it will no longer be valid and every API call using that API Key value will fail. When you decide to renew your API Key, please make sure that you use the new value in your BIP client."

kammerer:
@Nuanda Updated PR

Nuanda:
Pushed some small fixes. Otherwise :+1:

PR #245, opened 2015-05-08
shows pubmed link as another annotation in the popover

Nuanda:
This way we do not waste precious space in tables where 99% of records do not have a pubmed id.


kammerer:
:+1:

PR #246, opened 2015-05-08
Add more Search services specs

### LINKED ISSUE #196: Implement pending specs of Search class

kammerer:
Fixes #196 

I am not really sure how many different cases should we test here as there is a multitude of combinations possible. WDYT?

### DIFF
--- a/app/services/search.rb
+++ b/app/services/search.rb
@@ -3,7 +3,7 @@ class Search
   attr_accessor :query
 
   def initialize(query)
-    self.query = "*#{query}*"
+    self.query = query.include?("@") ? query : "*#{query}*"
   end
 
   def counts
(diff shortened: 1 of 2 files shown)


Nuanda:
Sure. Let's be pragmatic. We want to test searching for numeric values, like in the QTL table (a lot of numbers, but saved as text, sadly :( ). That is, it actually might be better for the search. So make sure this works.

The other case are email values in a current "annotation" field `entered_by_whom` - simply make sure the `at` character works (though I'm not sure we will index that field.

So make these two special cases work.

kammerer:
Added these two cases.

Nuanda:
Please fix the specs. Otherwise :+1:

kammerer:
You mean Travis? Not sure why it failed, now it passes (with the same seed on local too).

PR #250, opened 2015-05-08
Expose has_many associations in API responses

### LINKED ISSUE #216: Expose has_many associations in the API

kammerer:
Fixes #216


Nuanda:
:+1:

Nuanda:
BTW is that documented as well? That returned json will contain arrays of submodel ids?

PR #251, opened 2015-05-08
Non-flash csv export based on blob api

### LINKED ISSUE #46: Consider a non-flash fallback for csv export button

Nuanda:
Fixes #46


Nuanda:
I need that for browsers testing. Merging now and will address any issues risen by the review later.

PR #254, opened 2015-05-09
ES setup for five 'locus' CS model classes

Nuanda:
Tomasz, I did five further models. After you review that PR I will add all other models directly to master.


Nuanda:
Couldn't resists after I saw this: https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-model/examples/activerecord_associations.rb :). Provided we index all visible table columns and visible relations, the automatic json structure generator should work in most cases. In other cases (PP, LM, MA) one may redefine the structure.

Nuanda:
However, I've got a problem with travis stubbornly failing one of my tests (which work ok on my local machine).

Nuanda:
That is very strange - looks like Travis gets MapLocusHit model file from master, instead of this branch. What gives?..

Nuanda:
Ok, @kammerer sorry for a lot of spam. Still not sure what I'm doing wrong but please review it anyway - I'll sort the travis problem later, if it persists after master merge.

kammerer:
:+1:

kammerer:
I forgot one minor issue - new models should be added to `config/deploy.rb` to make sure indices are rebuild on deploy (also, if it takes to much time this code should be tweaked to rebuild only when necessary).

Nuanda:
Should those indices be built in same specific order?

kammerer:
It does not matter I think.

2015-05-11 18:35 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Should those indices be built in same specific order?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/254#issuecomment-100973097>
> .
>



-- 
Tomasz Szymczyszyn

PR #266, opened 2015-05-11
Escape special chars in search query

### LINKED ISSUE #247: Handle ES special characters / syntax in search query
The user input we collect in search gets passed to ES with small modifications only. This may cause ES to fail in case user enters something which ES recognizes as its own query string syntax but is not valid.

We should decide what power we give to users in terms of customizing search results, if any at all. The most straightforward option would be to strip / escape unwanted input. 

http://www.elastic.co/guide/en/elasticsearch/reference/1.x/query-dsl-query-string-query.html

kammerer:
Fixes #247

### DIFF
--- a/app/services/search.rb
+++ b/app/services/search.rb
@@ -3,7 +3,7 @@ class Search
   attr_accessor :query
 
   def initialize(query)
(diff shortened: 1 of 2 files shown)


Nuanda:
:+1:

PR #268, opened 2015-05-12
Use select2 for taxonomy term selects to maintain consistent looks

### LINKED ISSUE #259: Submissions: TaxonomyTerm select has different (browser default) style
Spotted in FireFox, on Linux.

kammerer:
Fixes #259


Nuanda:
Have you checked other submission selects for that case as well?

kammerer:
Yes, other are already using select2.

Nuanda:
:+1:

PR #271, opened 2015-05-13
Added fetch to filter info and disabled has_0 related buttons

### LINKED ISSUE #265: Datatable view improvements
I want to suggest two improvements for data-table.

1. Showing search query along with search results would make it clear what is going on.
2. "Related" button could be disabled if there are no related object so that user knows it is not worth clicking.

datatable
 view

Nuanda:
Fixes #265


kammerer:
:+1:

PR #272, opened 2015-05-13
Add uniqueness constraints to selected model attributes

### LINKED ISSUE #270: Add uniqueness vaildations for model classes
Important for issue #218

nowakowski:
Fixes #270

### DIFF
--- a/app/models/plant_population.rb
+++ b/app/models/plant_population.rb
@@ -25,6 +25,7 @@ class PlantPopulation < ActiveRecord::Base
 
   validates :name,
             presence: true,
+            uniqueness: true,
             allow_blank: true
 
   scope :by_name, -> { order('plant_populations.name') }
--- a/app/models/plant_population.rb
+++ b/app/models/plant_population.rb
@@ -25,8 +25,7 @@ class PlantPopulation < ActiveRecord::Base
 
   validates :name,
             presence: true,
-            uniqueness: true,
-            allow_blank: true
+            uniqueness: true
 
   scope :by_name, -> { order('plant_populations.name') }
 
--- a/app/models/plant_variety.rb
+++ b/app/models/plant_variety.rb
@@ -11,6 +11,7 @@ class PlantVariety < ActiveRecord::Base
   has_many :plant_lines
 
   validates :plant_variety_name,
+            presence: true,
             uniqueness: true
 
   include Filterable
--- a/app/models/plant_variety.rb
+++ b/app/models/plant_variety.rb
@@ -10,6 +10,9 @@ class PlantVariety < ActiveRecord::Base
 
   has_many :plant_lines
 
+  validates :plant_variety_name,
+            uniqueness: true
+
   include Filterable
   include Pluckable
   include Searchable
--- a/spec/factories/design_factor.rb
+++ b/spec/factories/design_factor.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :design_factor do
-    sequence(:design_factor_name) {|n| "#{Faker::Lorem.characters(5)}_#{n}"}
+    sequence(:design_factor_name) {|n| "#{Faker::Lorem.characters(20)}_#{n}"}
     institute_id { Faker::Company.name }
     trial_location_name { Faker::Lorem.sentence }
     design_unit_counter { Faker::Number.number(3).to_s }
--- a/spec/factories/design_factor.rb
+++ b/spec/factories/design_factor.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :design_factor do
-    sequence(:design_factor_name) {|n| "#{Faker::Lorem.characters(20)}_#{n}"}
+    sequence(:design_factor_name) {|n| "#{Faker::Lorem.characters(5)}_#{n}"}
     institute_id { Faker::Company.name }
     trial_location_name { Faker::Lorem.sentence }
     design_unit_counter { Faker::Number.number(3).to_s }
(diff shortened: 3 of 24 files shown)


nowakowski:
All existing CropstoreDB data has been verified to comply with the new validations.

app/models/plant_population.rb
```diff
@@ -25,6 +25,7 @@ class PlantPopulation < ActiveRecord::Base
 
   validates :name,
             presence: true,
+            uniqueness: true,
             allow_blank: true
```
Nuanda: Does uniqueness and allow_blank make any sense together? This just means there could be a single blank PP.name, right?

app/models/plant_variety.rb
```diff
@@ -10,6 +10,9 @@ class PlantVariety < ActiveRecord::Base
 
   has_many :plant_lines
 
+  validates :plant_variety_name,
+            uniqueness: true
```
Nuanda: PV.plant_variety_name shouldn't be required as well?

spec/factories/design_factor.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :design_factor do
-    sequence(:design_factor_name) {|n| "#{Faker::Lorem.characters(5)}_#{n}"}
+    sequence(:design_factor_name) {|n| "#{Faker::Lorem.characters(20)}_#{n}"}
```
Nuanda: You don't have to do that. If you use the sequence's "n" inside the value, factory girl should guarantee uniqueness. So 5 is enough.

spec/factories/linkage_group.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :linkage_group do
-    sequence(:linkage_group_label) { Faker::Lorem.characters(10) }
+    sequence(:linkage_group_label) { Faker::Lorem.characters(20) }
```
Nuanda: Instead of extending the string please use the sequence "n" value, as it is done in design_factor factory. Then, it is enough to have it shorter.

spec/factories/linkage_map.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :linkage_map do
-    sequence(:linkage_map_label) { Faker::Lorem.characters(10) }
+    sequence(:linkage_map_label) { Faker::Lorem.characters(20) }
```
Nuanda: Same comment as for LG.

spec/factories/plant_trial.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :plant_trial do
-    sequence(:plant_trial_name) {|n| "#{Faker::Lorem.characters(5)}_#{n}"}
+    sequence(:plant_trial_name) {|n| "#{Faker::Lorem.characters(20)}_#{n}"}
```
Nuanda: Again, I believe you don't have to do this, if you use "n". Apart from the case that you need it to be 20+ characters for some other reason.

spec/factories/plant_population.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :plant_population do
-    name { Faker::Lorem.word }
+    name { Faker::Lorem.characters(20) }
```
Nuanda: If name is required to be unique now, please use the sequence.

spec/services/search_spec.rb
```diff
@@ -67,11 +66,9 @@
...
-    lm1 = create(:linkage_map, linkage_map_label: 'linkage map label',
-                               map_version_no: '1',
+    lm1 = create(:linkage_map, map_version_no: '1',
                                plant_population: pp1)
-    lm2 = create(:linkage_map, linkage_map_label: 'linkage map label',
-                               map_version_no: '333',
+    lm2 = create(:linkage_map, map_version_no: '333',
                                plant_population: pp1)
     mp1 = create(:map_position, map_position: '102.8',
                                 linkage_group: lg1,
```
Nuanda: Why have you changed that file? Were these object violating new uniqueness constraints?

app/models/plant_population.rb
kammerer: Event without uniqueness, does allow_blank make any sense at all?

nowakowski: It would allow any number of NULL records - however you are right in that it doesn't make much sense. I've checked and this field is not blank in any existing PP records; accordingly, I will remove this clause.

app/models/plant_variety.rb
nowakowski: Okay.

spec/factories/design_factor.rb
nowakowski: I kept getting transient Fs due to validation violations. Let me check.

spec/factories/linkage_group.rb
nowakowski: Good idea.

spec/services/search_spec.rb
nowakowski: Indeed they were - in most `search_spec` specs.

Nuanda: There is something odd, then. You see there are 2-3 objects per each model and they usually have non-unique attributes assigned to them (that is actually kind of point here, as we want to test searching).

Anyway, removing parameters as you did here is not a good solution for that test. If it keeps giving you failures after the factory changes, I'll try to look into it.

Nuanda:
:+1:

PR #273, opened 2015-05-13
Refactor submissions javascript

kammerer:
Small refactoring - basically split larger methods into smaller ones.


Nuanda:
Made a quick test and seems to be (still) working :). :+1:

PR #274, opened 2015-05-13
Different submission buttons

Nuanda:
This is my fault but I don't like the way current buttons are presented to the user during submission steps. So I decided to go back to @bchucherko design and arrange them exactly that way.

https://live.uxpin.com/3d91dec11a9043767b221f04117dcdfd221c864a#/pages/19058672

It's an improvement, in my opinion.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -33,12 +33,17 @@ class SubmissionsController < ApplicationController
       redirect_to edit_submission_path(@submission) and return
     end
 
-    unless @content.valid?
+    unless @content.valid? || params[:leave]
       render action: :edit and return
     end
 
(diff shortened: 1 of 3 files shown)


Nuanda:
Tomasz, feel free to change it any way you like.

kammerer:
The only problem I see is that it does not actually save changes from the current step.

Nuanda:
@kammerer This is a very good point. Problem is with failed validations - it would be weird to force the user to mend invalid form parts, when (s)he wants to leave. What I propose is to have the "save" part of that action do sth like _save as much as you can without bothering the user_, and then we'll still have the user return to the same form later, so nothing's should be lost.

kammerer:
:+1:

PR #281, opened 2015-05-15
Create API

### LINKED ISSUE #218: Implement create API
Validation should handle misnamed attributes.

kammerer:
Fixes #218 

Create API allows to set all attributes except for `id`, `user_id`, `created_at` and `updated_at` which are blacklisted. This includes FKs for `belongs_to` relationships.


kammerer:
API docs are somewhat improved but attributes' lists are still incomplete.

config/routes.rb
```diff
@@ -30,16 +30,13 @@
     namespace :v1 do
       resources :resources, except: [:new]
 
-      Brassica::Api.readable_models.map { |klass| klass.name.underscore.pluralize }.each do |model_name|
+      Api.readable_models.map { |klass| klass.name.underscore.pluralize }.each do |model_name|
```
Nuanda: Will that make Qtl into qtls? If so, do we want it?

app/views/application/api/_index.html.haml
```diff
@@ -0,0 +1,20 @@
+.action
+  %span.method GET
+  %span.url= "/api/v1/#{model.name.underscore.pluralize}"
+
+= api_props 'Search params', "#{model.name.underscore}.index.params"
+
+%h4 Example
+%pre
+  %code.http= t("api.#{model.name.underscore}.show.example", url: "/api/v1/#{model.name.underscore.pluralize}?per_page=3")
```
Nuanda: Should be .index.example, I think.

app/controllers/api/v1/resources_controller.rb
```diff
@@ -4,51 +4,91 @@ class Api::V1::ResourcesController < ApplicationController
 
   before_filter :authenticate_api_key!
   before_filter :require_allowed_model
-
-  # FIXME find a way to document api calls
+  before_filter :require_strictly_correct_params, only: :create
 
   def index
-    filter_params = params[model_name.singularize].presence
+    filter_params = params[model_name].presence
```
Nuanda: So it would be consistent with datatables calls, that use plural names, right?

app/controllers/api/v1/resources_controller.rb
```diff
@@ -4,51 +4,91 @@ class Api::V1::ResourcesController < ApplicationController
...
+    end
   end
 
   private
 
   def authenticate_api_key!
-    token = ApiKey.normalize_token(params[:api_key])
-    unless token.present? && ApiKey.exists?(token: params[:api_key])
+    unless api_key.present?
       render text: "Not Found", status: 404
```
Nuanda: Can we add a longer, more informative message for the user here?

app/controllers/api/v1/resources_controller.rb
```diff
@@ -4,51 +4,91 @@ class Api::V1::ResourcesController < ApplicationController
...
+    token = params[:api_key] || request.headers["X-BIP-Api-Key"]
+    token = ApiKey.normalize_token(token)
+
+    @api_key = token && ApiKey.find_by(token: token)
+  end
+
   def require_allowed_model
-    unless allowed_models.include?(model_name)
+    if request.request_method_symbol == :get && !Api.readable_model?(model_name)
+      raise ActionController::RoutingError.new('Not Found')
```
Nuanda: Why sometimes we throw a RoutingError (that, I assume, results in a 404) and sometime (the no key case handling above) we simply render text with status = 404?

app/controllers/api/v1/resources_controller.rb
```diff
@@ -4,51 +4,91 @@ class Api::V1::ResourcesController < ApplicationController
...
+  def require_strictly_correct_params
+    model_attrs = model_klass.attribute_names
+    misnamed_attrs = (params[model_name].try(:keys) || []) - model_attrs
+
+    if misnamed_attrs.present?
+      errors = []
+
+      misnamed_attrs.each do |attr|
+        errors << { attribute: attr, message: "Unrecognized attribute name" }
+      end
```
Nuanda: Of course it's a matter of taste, but I guess one could write `errors = misnamed_attrs.map do ...` here.

app/controllers/api/v1/resources_controller.rb
```diff
@@ -4,51 +4,91 @@ class Api::V1::ResourcesController < ApplicationController
...
+
+      render json: { errors: errors }, status: 422
+    end
+    true
+  end
+
   def model_name
-    @model_name ||= request.path.match(/\A\/api\/v1\/(([\w_]+)\/?)/)[2]
+    @model_name ||= request.path.match(/\A\/api\/v1\/(([\w_]+)\/?)/)[2].singularize
```
Nuanda: Is it possible to have it mapped directly in routes.rb - sth like /api/v1/:model_name/... that would turn what's on the :model_name position into a new param?

config/routes.rb
kammerer: I think we want that for consistency in the API. Actual naming of db table does not matter.

app/controllers/api/v1/resources_controller.rb
kammerer: Hmm, no idea why I used raise, IIRC render in before filter should work as well.

app/controllers/api/v1/resources_controller.rb
kammerer: I haven't thought of this but what you suggest with a `:constraint` option would work.

app/views/application/api.html.haml
```diff
@@ -18,7 +18,7 @@
       %p
         BIP provides HTTP REST-style API which uses JSON to represent returned
         resources. Full access to the API is allowed for all registered users.
-        In order to authenticate users pass <code>api_key</code> parameter in
+        In order to authenticate users pass <code>X-BIP-Api-Key</code> header in
```
Nuanda: I think the controller part does still not exclude passing api key as a param. But I understand we want to encourage passing it in a header?

config/locales/api.en.yml
```diff
@@ -25,7 +62,16 @@ en:
           format: int
           desc: "Total number of existing records"
 
+      request: |-
+        GET /api/v1/plant_lines?page=2&per_page=100 HTTP/1.1
+        Content-Type: application/json
+        Accept: application/json
+        X-BIP-Api-Key: ba6d4625927ede0db8768782994d9276cfff9ba79e8df8c831aedc453e850836
```
Nuanda: I understand you used user's own api key to construct sample api calls in the docs view. Why change to predefined 'demo' keys now?

spec/models/model_spec.rb
```diff
@@ -87,7 +87,7 @@
   end
 
   context "supporting API" do
-    Brassica::Api.readable_models.each do |klass|
+    Api.readable_models.each do |klass|
       context klass do
         it "includes Filterable" do
           expect(klass.ancestors).to include(Filterable)
```
Nuanda: One additional request - I think there is value in providing read only access to tables like TaxonomyTerm, that do not support filtering. It kind of works out of the box (I just tested it and both index and show do). Can we exclude them from the requirement of being Filterable (as long as they do not list query/search/fetch in their props/params)?

config/routes.rb
Nuanda: All right.

config/locales/api.en.yml
kammerer: Do you think it is confusing that the key shown here is not the real one used in sample api calls? I didn't want to expose user's key here so that user does not expose it by mistake (e.g. copy and paste somewhere).

kammerer: Maybe I should use a special predefined key with limited capabilities instead of user's own key?

spec/models/model_spec.rb
kammerer: No problem.

config/locales/api.en.yml
Nuanda: I think it's all right as it is, with GET APIs. Just wanted to know what was the rationale.

Nuanda:
Regarding the attribute list. What would you say for a further model evolution, to have permitted (sort of "by default") all visible (i.e. table_columns) parameters of the given model (i.e. those that are not joined in from related models)?

Currently I was adding permitted parameters whenever I needed them in datatables links. But for API, it kind of feels strange to have 2 chosen parameters filterable and 3 other blocked...

app/controllers/api/v1/resources_controller.rb
kammerer: No, it is not consistent but datatables controller is not public API. `model_name` here is singular and I think it is more natural and consistent with params for create action.

Nuanda: All right.

kammerer:
Updated PR. I am OK with allowing more filterable parameters.

Nuanda:
:+1:

PR #288, opened 2015-05-21
Improvements in api

Nuanda:
A lot of things:
 - added all models that should be readable to API
 - moved examples and several attributes to default section
 - support for excluding some fields and expanding some related objects in json
 - a lot of field descriptions.


Nuanda:
@kammerer Tomasz, since I want to deploy it with other changes, I'll merge this PR without review. But I'll leave it here in case you'd like to go back to it and check the changes.

PR #291, opened 2015-05-22
Create api improvements

Nuanda:
Some further improvements in the create API:
 - switched from session auth to key based auth
 - returns json instead of text for 40x messages
 - rebuilt api docs with separate section on data submission
 - other smaller things.


kammerer:
:+1:

PR #292, opened 2015-05-25
Error mailing

### LINKED ISSUE #61: Add automatic error reporting to admins via email

Nuanda:
Fixes #61 

Cannot test on  production at the moment since the port to the mail server is blocked.


kammerer:
What about ActionMailer configuration?

Nuanda:
It kind of looks like an overkill to me - see http://api.rubyonrails.org/classes/ActionMailer/Base.html. Perhaps, if we were sending mails to user. But just to report an exception - I don't know.

Also, exception notifier works well for us in many previous projects.

kammerer:
I was just asking how does it actually send these emails. I mean,
`exception_notification` depends on `action_mailer` to do so and I think it
does not work out of the box (it did not on my box).

2015-06-01 10:41 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> It kind of looks like an overkill to me - see
> http://api.rubyonrails.org/classes/ActionMailer/Base.html. Perhaps, if we
> were sending mails to user. But just to report an exception - I don't know.
>
> Also, exception notifier works well for us in many previous projects.
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/292#issuecomment-107366770>
> .
>



-- 
Tomasz Szymczyszyn

Nuanda:
To make it work locally, you need to copy mailer config to development.rb. You might need to change `sender_address` to `yourlogin@localhost` or something. Now, such email will be ignored by some mail servers, but gmail is tolerant enough to put it in your spam folder.

You can change the destination email in your `.env` file.

Nuanda:
BTW it uses whatever is found under your `sendmail` executable alias.

kammerer:
I am confused, there is no mailer config for any env. Sure I could add it but shouldn't it be included in the PR?

Nuanda:
It is in `config/environments/production.rb`, commited to this PR.

kammerer:
This is `exception_notification` config but AFAIK it does not know how to
send emails itself and uses `action_mailer` to do that. Isn't that right?
README says it should be configured:
https://github.com/smartinez87/exception_notification#email-notifier

2015-06-01 17:05 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> It is in config/environments/production.rb, commited to this PR.
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/292#issuecomment-107563196>
> .
>



-- 
Tomasz Szymczyszyn

Nuanda:
Tomasz, I have added the mailer options as you requested. Still timeout error on production, which is probably due to us being on a dev/testing server. Will address the issue again when we go public.

kammerer:
The code looks good, still I am not able to make it work on my local box :( I tried sendmail from CLI and it works fine.

Nuanda:
Ok, lets leave it for now. When we get the production server I'll try to work with TGAC admins to make it work there. Then, we will merge this PR with master.

kammerer:
:+1:

PR #294, opened 2015-05-25
Refactor linkage groups

### LINKED ISSUE #207: Nature of LinkageGroup - LinkageMap relation
Every LinkageGroup in current data has exactly 1 LinkageMap related to it. However, the schema relation between these two tables is a "many-to-many" relation.

Question: should this relation be reduced to a "many-to-one" relation or is it just a peculiar case of the current CS data and, in the future, any LinkageGroup should be allowed to have many LinkageMaps related to it?

nowakowski:
Fixes #207


nowakowski:
@Nuanda could you test impact on GUI views?

Nuanda:
:+1:

PR #296, opened 2015-05-29
Rolling submission back to specific step

### LINKED ISSUE #285: Handle new plant line name uniqueness validation on submission finalization
Happens in a very rare case when two simultaneous PP submissions are in place, in their 4th stage both, and both having defined a new PL with exactly the same name.

Submitting the first goes ok, as it should, and submitting the second fails on the PL name validation uniqueness.

Nuanda:
Fixes #285


kammerer:
:+1:

PR #303, opened 2015-06-02
Do not wildcard queries for numeric fields of Qtl model

### LINKED ISSUE #253: Do not wildcard numeric fields in global search
Example: `map_locus.map_locus = "102.5"`

This should come up in search by "102.5" but it should not when one looks for "2.5" or "2". I can supply you with a set of fields that should be treated this way.

Wait for #242 before doing that. Also related to #225.

kammerer:
Fixes #253. Should be pretty easy to implement for other models. ES index needs to be rebuilt to check how it works.

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -12,6 +12,10 @@ module Searchable extend ActiveSupport::Concern
 
     after_touch { __elasticsearch__.index_document }
 
+    def self.numeric_columns
+      []
+    end
+
     def self.indexed_json_structure
       return @indexed_json_structure if @indexed_json_structure
       only = table_columns.select{ |c| !c.include? '.' }.map(&:to_sym)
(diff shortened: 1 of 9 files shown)


Nuanda:
Tested, works for me. I will add further fields to treat them this way in the issue discussion.

kammerer:
Updated models searchable by `map_positions.map_position`.

app/models/concerns/searchable.rb
```diff
@@ -30,4 +34,10 @@ def as_indexed_json(options = {})
       as_json(self.class.indexed_json_structure)
     end
   end
+
+  def self.classes
+    ActiveRecord::Base.descendants.select { |klass|
+      klass.ancestors.include?(Searchable)
+    }
```
Nuanda: Let's use do/end for multiline blocks, and reserve multiline { } use only for hash structures. Ok?

kammerer: Fixed

Nuanda:
:+1:

PR #305, opened 2015-06-03
Use ActionController::Metal for API

kammerer:


Nuanda:
This Travis failure seems important, please investigate.

kammerer:
Forgot to add new file...

Nuanda:
Still some (probably) random spec fail, but :+1:

PR #312, opened 2015-06-15
Allow to reach search results view with an url

kammerer:


app/assets/javascripts/search.coffee
```diff
@@ -38,9 +43,14 @@ class Search
...
         @$results.html(response)
+        @updateHistory(term)
 
-      complete:
+      complete: =>
         @ajax = null
 
+  updateHistory: (term) =>
+    if window.location.search != "?search=#{term}"
+      window?.history?.pushState(null, null, "/?search=#{term}")
```
Nuanda: Wow, didn't know about that! This is serious business :).

Nuanda:
Tested, works fine. :+1:

PR #322, opened 2015-06-23
Api for plant population list

### LINKED ISSUE #318: Enable PlantPopulationList as an API-writable resource
When submitting a PP, let the user to set plant_lines_ids, and the other way around. We don't need to support other _ids to-many relation, since they may be set on the other side (with the _id field), but we need to support n-to-m relationships.

### LINKED ISSUE #319: Plenty of problematic defaults in database model - please remove them
Piotr, I still find a lot of DB default settings to "unspecified". I guess these are still remnants of wrong CropStore design lingering on in the current BIP model.

Please remove them all, they are of no use. If we do not require a certain field set, null value is ok. If we do require it, having a default only confuses the user who thinks the filed is not really required after all (the DB takes care of it by putting the default value in).

### LINKED ISSUE #320: PPL sort_order - what is the meaning of that field
Piotr, I need to introduce PPL as an API-accessible and API-submittable model. So I need to document its fields. Can you tell me what is the meaning of the `sort_order` filed?

Or, if it is not needed, could be simply remove that column from the model?

I also think that perhaps we do not need the annotation columns there (comments, entered_by_whom, date_entered, data_provenance, confirmed_by_whom) - after all, it is essentially just an N-to-M table...

### LINKED ISSUE #321: Change HABTM relation of PP and PL into HMT relation
PPL needs to become an addressable entity since it is going to be an API resource. Therefore we need it to have its own :id attribute. I guess this requires the nature of n-to-m relation between PP and PL be changed from HABTM to the "through" type of relation.

Also please make sure that when I remove a PL, all related PPs stay intact, but all records from PPL that linked that PL to any PPs are removed.

Nuanda:
When finished, it should:

Fixes #321 

Fixes #318 

Fixes #319


Nuanda:
@nowakowski Piotr, this partially fixes #321 now - changes habtm to hmt between PPs and PLs. Please take case for the second part of that issue - proper cascade when on side of the relation is destroyed.

Also this needs #320 to be addressed.

Nuanda:
@nowakowski Ok, I added FKey declarations for user relation as well. Please fix the second part of #321 (I believe there is a failing spec about that already).

Nuanda:
#320 stays not fixed but we may merge it as it is.

Issue #314, opened 2015-06-22
Classify all BIP table relations as either mandatory or optional


nowakowski:
The following is a dump of all BIP model tables and relation columns (regular attributes have been omitted). For each relation column, please provide information on whether the relevant relation is mandatory (i.e. a child object *must* have a parent object assigned) or optional (a child object can be valid even when the parent object is NULL).

```
# encoding: UTF-8
# This file is auto-generated from the current state of the database. Instead
# of editing this file, please use the migrations feature of Active Record to
# incrementally modify your database, and then regenerate this schema definition.
#
# Note that this schema.rb definition is the authoritative source for your
# database schema. If you need to create the application database on another
# system, you should be using db:schema:load, not running all the migrations
# from scratch. The latter is a flawed and unsustainable approach (the more migrations
# you'll amass, the slower it'll run and the greater likelihood for issues).
#
# It's strongly recommended that you check this file into your version control system.

ActiveRecord::Schema.define(version: 20150608102750) do

  create_table "countries", force: :cascade do |t|
  end

  create_table "design_factors", force: :cascade do |t|
  end

  create_table "genotype_matrices", force: :cascade do |t|
    t.integer "linkage_map_id"
  end

  create_table "linkage_groups", force: :cascade do |t|
    t.integer "linkage_map_id"
  end

  create_table "linkage_maps", force: :cascade do |t|
    t.integer "plant_population_id"
  end

  create_table "map_locus_hits", force: :cascade do |t|
    t.integer "linkage_map_id"
    t.integer "linkage_group_id"
    t.integer "population_locus_id"
    t.integer "map_position_id"
  end

  create_table "map_positions", force: :cascade do |t|
    t.integer "linkage_group_id"
    t.integer "population_locus_id"
  end

  create_table "marker_assays", force: :cascade do |t|
    t.integer "restriction_enzyme_a_id"
    t.integer "marker_sequence_assignment_id"
    t.integer "restriction_enzyme_b_id"
    t.integer "primer_a_id"
    t.integer "primer_b_id"
    t.integer "probe_id"
  end

  create_table "marker_sequence_assignments", force: :cascade do |t|
  end

  create_table "plant_accessions", force: :cascade do |t|
    t.integer "plant_line_id"
  end

  create_table "plant_lines", force: :cascade do |t|
    t.integer  "taxonomy_term_id"
    t.integer  "plant_variety_id"
  end

  create_table "plant_parts", force: :cascade do |t|
  end

[...]

PR #323, opened 2015-06-24
Added foreign key declarations and purged obsolete field in trait_scores

### LINKED ISSUE #317: Add foreign key declarations in BIP model to guard against data inconsistency
See http://edgeguides.rubyonrails.org/4_2_release_notes.html#foreign-key-support

nowakowski:
Fixes #317

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,7 @@ group :development, :test do
   gem 'pry-rails'
   gem 'pry-doc'
   gem 'pry-nav'
+  gem 'thin'
 end
 
 group :test do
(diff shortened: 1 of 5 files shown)


nowakowski:
**Caution** This operation is somewhat premature as long as issue #314 remains open. For the time being I have set the most liberal enforcement rules possible for all FKs (set null on delete, cascade on update). Any relation classified as mandatory in #314 will at least need to have its associated deletion rule upgraded to :cascade (or possibly :restrict -- requiring further changes in `spec/models/model_spec.rb`).

Nuanda:
I rather think :restrict, but we will deal with that later, when we add relation existence requirement.

I understand that all these FKs allow to set nil value, and they set nil for related_id values when the related resource is being destroyed?

nowakowski:
Yes, that's the meaning of `:nullify`.

Nuanda:
:+1:

PR #328, opened 2015-06-26
API for setting HABTM PV-Country relations

### LINKED ISSUE #324: Solve setting PV countries relations by API
There are two habtm relations between PV and Countries table. Since PV is API-writable now, please make sure there is a way for the user to set these relations.

kammerer:
Fixes #324


spec/services/api/create_params_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'rails_helper'
+
+RSpec.describe Api::CreateParams do
+  describe "#permissions" do
+    it "allows scalar model attributes" do
+      pl_permissions = Api::CreateParams.new(Api::Model.new('plant_line'), {}).permissions
+      # FIXME plant_variety_name - shouldn't it be removed?
```
Nuanda: What do you mean by that?

kammerer: PL belongs to PV and also has `plant_variety_name` attribute. Isn't that
redundant?

2015-06-26 12:18 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> In spec/services/api/create_params_spec.rb
> <https://github.com/eSpectrum-IT/brassica/pull/328#discussion_r33344083>:
>
> > @@ -0,0 +1,53 @@
> > +require 'rails_helper'
> > +
> > +RSpec.describe Api::CreateParams do
> > +  describe "#permissions" do
> > +    it "allows scalar model attributes" do
> > +      pl_permissions = Api::CreateParams.new(Api::Model.new('plant_line'), {}).permissions
> > +      # FIXME plant_variety_name - shouldn't it be removed?
>
> What do you mean by that?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/328/files#r33344083>.
>



-- 
Tomasz Szymczyszyn

Nuanda: Ah, all right, I though this is about PV. Well, I guess this is one of several "leave it for later curation" cases - perhaps some PLs couldn't be mapped to PVs so Wiktor decided it would be safe to leave that attribute.

So removing that from the DB is rather out of question (or am I wrong - @nowakowski Piotr?). In such a case I vote we shouldn't hide it in API either, though it looks kind of strange.

Nuanda:
Out of curiosity - why separate Model and AssociationFinder? The later looks like a bunch of Model methods to me.

spec/services/api/create_params_spec.rb
kammerer: M'key, I will edit the note then (if Piotr confirms). Don't you think it would be good to put these things "left for later curation" somewhere (docs)?

Nuanda: Well, they are "documented" - in past GitHub issues. I leave it up to Piotr if and how he'd like to maintain a separate "curation pending" list.

kammerer:
No particular reason, merging them seems to be a good idea.

spec/services/api/create_params_spec.rb
nowakowski: Unfortunately you're correct - here's your problem:
<pre>
bip_development=# SELECT COUNT(*) FROM plant_lines WHERE plant_variety_name NOT IN (SELECT plant_variety_name FROM plant_varieties);
count 
-------
 15626
(1 row)
</pre>
It is safe to omit this column in all new `PlantLine` records (and thus not include it in any input forms) but existing data should retain it.

kammerer:
Merged `Api::Model` and `Api::AssociationFinder`. I will create a new issue for adding `PlantLine#plant_variety_name' to the API.

Nuanda:
Tomasz, please also expose Countries table as readonly API resource. Otherwise it will be difficult for users to find required country ids.

kammerer:
Added Country to read API.

Nuanda:
:+1:

PR #355, opened 2015-07-10
PT submission: erase all step03 data when re-done step02

### LINKED ISSUE #340: PT submission: erase all step03 data when re-done step02
Since when a user goes back from step04 to step02 and changes something in TraitDescriptors we have no idea if TraitScores recorded in step03 are still valid, please remove all step03 content (when user goes again from step02 to step03) and force the user to reupload the scores file.

Step04 content may stay as it is.

kammerer:
Fixes #340

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -104,8 +104,8 @@ class Submission < ActiveRecord::Base
     step02_was = content_was[:step02] || {}
     step02 = content[:step02] || {}
 
-    old_trait_descriptor_list = (step02_was[:trait_descriptor_list] || []).map(&:to_s).sort
-    new_trait_descriptor_list = (step02[:trait_descriptor_list] || []).map(&:to_s).sort
+    old_trait_descriptor_list = (step02_was[:trait_descriptor_list] || []).map(&:to_s)
+    new_trait_descriptor_list = (step02[:trait_descriptor_list] || []).map(&:to_s)
 
     if old_trait_descriptor_list != new_trait_descriptor_list
       content.clear(:step03)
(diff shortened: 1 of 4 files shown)


app/models/submission/content.rb
```diff
@@ -17,6 +17,13 @@ def update(step, step_content)
       merge(step => current_step_content.merge(step_content))
   end
 
+  def clear(step)
+    raise Submission::InvalidStep, "No step #{step}" unless submission.steps.include?(step.to_s)
+    submission.content = submission.
+      read_attribute(:content).
```
Nuanda: read_attribute is here not to cause nil.merge?

app/models/submission.rb
```diff
@@ -96,6 +97,21 @@ def set_defaults
...
 
+  def apply_content_adjustments
+    return unless trial?
+    return unless content_changed?
+
+    step02_was = content_was[:step02] || {}
+    step02 = content[:step02] || {}
+
+    old_trait_descriptor_list = (step02_was[:trait_descriptor_list] || []).map(&:to_s).sort
+    new_trait_descriptor_list = (step02[:trait_descriptor_list] || []).map(&:to_s).sort
```
Nuanda: Please compare without sorting. Due to problems with possible name duplicates, I had to change the semantics of trait descriptor list and now the order of contents is meaningful. Therefore, simply changing places of same two traits in the list should also delete step03 content.

app/models/submission/content.rb
kammerer: No, it is here because `Submission#content` returns a `Submission::Content` instance and I want a 'raw' hash. There should be no `nil` here as a default is set before submission is created.

app/models/submission.rb
kammerer: Done.

Nuanda:
:+1:

PR #358, opened 2015-07-22
Fix reappearing TDs and PLs in submission

### LINKED ISSUE #353: New TDs appear in the list after being removed
1. Add new TD
2. Go to step03
3. Go back to step02
4. Remove new TD and go to step03
5. Go back to step02 - new TD is visible in the list (it is not saved in trait_descriptor_list but it is in new_trait_descriptors)

kammerer:
Fixes #353


Nuanda:
:+1:

PR #359, opened 2015-07-22
Fix redirect loop triggered by clicking a link after session timeout

### LINKED ISSUE #354: Redirect loop after session expiration
When I leave my browser open and the session cookie expires I get a message about redirect loop from chrome upon clicking anything.

kammerer:
Fixes #354


Nuanda:
Apart from my submissions this is probably the only place where redirecting to the front page was strange. So it's fine. :+1:

PR #360, opened 2015-07-22
Show additional information in autocomplete result lists

kammerer:
Slightly enhance (I think) autocomplete fields used to select PP, PL, PV by showing more information for each entry.


Nuanda:
Checked for PlantLine and PlantVariety, works ok. :+1:

PR #361, opened 2015-07-23
Single row combo field

kammerer:
Makes combo fields appear in a single form row (on big screens). The forms are rather big and this makes them slightly more compact and spares some scrolling.


Nuanda:
:+1:

PR #370, opened 2015-07-24
better annotations form wrapping for smooth animation

### LINKED ISSUE #365: Additional annotations in new PL/TD subform expand/collapse with a glitch in animation
Most likely due to vertical margins collapsing.

Nuanda:
Tested in FF, Opera, Chromium. Kind of tested in Safari (due to remote VM access lag only made sure the animation collapse work, but the refresh rate I have now is too poor to really check for anomation smoothness).

Fixes #365 

And yes - it was connected to http://bugs.jquery.com/ticket/7442

### DIFF
--- a/app/assets/stylesheets/partials/_submissions.scss
+++ b/app/assets/stylesheets/partials/_submissions.scss
@@ -36,6 +36,10 @@
     margin-top: 4em;
   }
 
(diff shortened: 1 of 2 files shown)


kammerer:
:+1:

PR #373, opened 2015-07-24
fixed search history management for safari

### LINKED ISSUE #357: Needed slightly more versatile history management code
@kammerer Tomasz, I needed to block search-history management javascript code for now, since it caused Safari to reload the page indefinitely (see previous bug filed by Wiktor and may latest commit to master).

If you could reinstate that code in a way that does not cause problems on Safari, it would be great.

Nuanda:
Fixes #357 

Done by assigning a state object to the pushState call and checking that object presence in the popstate handler. Also preserving old code for non-Safari browsers.

### DIFF
--- a/app/assets/javascripts/application.coffee
+++ b/app/assets/javascripts/application.coffee
@@ -38,7 +38,12 @@ $ ->
     bottom: ->
       return @bottom = - bh.top + 320
 
-#  # Handles navigation through history states originating from client-side
-#  # (e.g. search results)
-#  $(window).on 'popstate', ->
-#    window.location.replace window.location.href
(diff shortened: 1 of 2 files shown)


kammerer:
:+1: 

I am not sure why branching is necessary - I mean, `state` should be available in the event in all browsers, shouldn't it? But I checked it in chrome and it is null...

Nuanda:
Yes, there are some standard implementation problems, esp. in webkit browsers. Safari (faulty) fires that callback on page load. So some say state==null is enough, other say it is not. So branching is there to make it extra safe for non-Safari browsers.

PR #375, opened 2015-07-27
Simple rails cache for DT jsons

Nuanda:
Tomasz, this is a new PR to discuss and co-develop data table creation and transfer optimization.

What I pushed is a simple rails cache mechanism to store jsons ready to be shipped to the client. The cache key is based on entire model count and the model-wide latest updated_at, so it should be error-proof. It is valid infinitely so given appropriate auto-outdating strategy we may set it a very long expiration date (or do I miss sth important here?).

For all plant lines is makes >10x faster data retrieval and preparation on my localhost (this, however, does not tackle the problem of data transfer and rendering). What is nice is that is works for all clients (so coming second you already hit the cache on you first pare visit). Any submission and deletion should make the cache MISSED for the first time.

TODOs:
- [x] check what old cache purge strategy the in-memory rails cache has and if we need an external service (like memcached) for that or not
- [ ] after everything is done and tested, we may revert "perform_caching = true" in the development mode (thought it is not strictly required to remove it).

I'm keen to know you opinion about that (and what other optimization strategies you see fit for our case).

### DIFF
--- a/app/controllers/data_tables_controller.rb
+++ b/app/controllers/data_tables_controller.rb
@@ -6,9 +6,15 @@ class DataTablesController < ApplicationController
         model_param
       end
       format.json do
-        objects = model_param.singularize.camelize.constantize.table_data(params)
-        grid_data = ApplicationDecorator.decorate(objects)
-        render json: grid_data.as_grid_data
+        cache_key = params.reject{ |k,_| %w(_ controller action).include? k }
+        cache_key[:latest_change] = params[:model].singularize.camelize.constantize.maximum('updated_at')
+        cache_key[:count] = params[:model].singularize.camelize.constantize.count
+        logger.debug "CACHE KEY: #{cache_key}"
+        grid_data = Rails.cache.fetch(cache_key, expires_in: 300.days) do
+          logger.debug 'MISS MISS MISS'
+          prepare_grid_data.to_json
+        end
+        render json: grid_data
       end
     end
   end
@@ -19,6 +25,11 @@ class DataTablesController < ApplicationController
   end
 
   private
+
+  def prepare_grid_data
+    objects = model_param.singularize.camelize.constantize.table_data(params)
+    ApplicationDecorator.decorate(objects).as_grid_data
+  end
 
   def model_param
     if params[:model].present? && !allowed_models.include?(params[:model])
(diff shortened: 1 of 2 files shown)


kammerer:
I like this idea. I am not sure though if we should use the default file store or some kind of memory store. Anyway, we should monitor disk or memory, if it is not done yet with NAGIOS I can add checks to `monit` config. 

As for other optimizations, I've changed `nginx` config so that datatable responses are gzipped. It gives some performance improvement, though not as dramatic as caching.

Nuanda:
Good idea - even earlier today I was thinking that switching gzip may compress some json responses quite a bit.

Please add the memory and disk probes to the monitoring system. Still, I know, we have to wait for the mailing fix - but hopefully this will eventually be opened.

kammerer:
Added system level memory and disk checks, details are in the wiki. Depending on chosen caching store additional check preventing cache from growing too much may be needed.

Nuanda:
Default memory cache is per-process and I think we better keep to the disk cache. However, I learnt it has no expiration strategy - i.e. it grows until the disk is full and then fails. One should manually take care for older entries: http://api.rubyonrails.org/classes/ActiveSupport/Cache/FileStore.html#method-i-cleanup

If this is the case, perhaps we should cache only all-records queries, handling filtered queries with no cache. Some such queries will still be somewhat slower (for instance a trait_score query loading 10k+ scores) but otherwise it is not possible to estimate the cache size. Currently I cache by all filtering params.

Or am I exaggerating and we should not worry about the disk space?

kammerer:
Well, we have quite a lot of disk space available (~56GB), so maybe we
could cache everything and employ cron to perform cleanup?

2015-07-30 10:10 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Default memory cache is per-process and I think we better keep to the disk
> cache. However, I learnt it has no expiration strategy - i.e. it grows
> until the disk is full and then fails. One should manually take care for
> older entries:
> http://api.rubyonrails.org/classes/ActiveSupport/Cache/FileStore.html#method-i-cleanup
>
> If this is the case, perhaps we should cache only all-records queries,
> handling filtered queries with no cache. Some such queries will still be
> somewhat slower (for instance a trait_score query loading 10k+ scores) but
> otherwise it is not possible to estimate the cache size. Currently I cache
> by all filtering params.
>
> Or am I exaggerating and we should not worry about the disk space?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/375#issuecomment-126218897>
> .
>



-- 
Tomasz Szymczyszyn

PR #377, opened 2015-07-31
HTTP caching for datatables

### LINKED ISSUE #363: Enable HTTP caching for browsing data
Plant lines especially take quite a bit to load (a few seconds at least).

kammerer:
A few small changes were required to make this work:
1. By default datatables make js append cache buster to url params. Solution: add `cache: true` to datatables ajax options.
2. Browser tried to send Etag generated for HTML instead of JSON response. Solution: append format indicator to the url.

Etag generation is handled by RoR itself (AFAIK Etag is computed from rendered response), that's why no other changes in code are required.

Finally, caching will NOT work with invalid SSL certificate. I spend a while to discover this...

Fixes #363

### DIFF
--- a/app/helpers/data_tables_helper.rb
+++ b/app/helpers/data_tables_helper.rb
@@ -3,7 +3,8 @@ module DataTablesHelper
     data_tables_path(
       query: params[:query],
       fetch: params[:fetch],
-      model: model_param
+      model: model_param,
+      format: :json
     )
   end
 
@@ -12,7 +13,12 @@ module DataTablesHelper
       :table,
       class: 'table table-condensed table-browse table-hover data-table',
       id: model_param.dasherize,
-      data: { ajax: datatables_source }
+      data: {
+        ajax: {
+          url: datatables_source,
+          cache: true
+        }
+      }
     ) do
       content_tag(:thead) do
         content_tag(:tr) do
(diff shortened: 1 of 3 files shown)


kammerer:
Sorry, need to fix spec yet.

app/helpers/data_tables_helper.rb
```diff
@@ -3,7 +3,8 @@ def datatables_source
     data_tables_path(
       query: params[:query],
       fetch: params[:fetch],
-      model: model_param
+      model: model_param,
+      format: :json
```
Nuanda: You may want to remove the new param from cache key in the index operation - though I guess this is not needed, since it does not change anyway.

Nuanda:
How can I test it with no valid ssl certificate?

kammerer:
You need to trust this cert or add it to exceptions' list (e.g. in Firefox).

2015-07-31 16:07 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> How can I test it with no valid ssl certificate?
>
> —
> <https://github.com/eSpectrum-IT/brassica/pull/377#issuecomment-126703659>
> .
>



-- 
Tomasz Szymczyszyn

Nuanda:
:+1:

PR #425, opened 2016-03-11
Add published flag to model tables

### LINKED ISSUE #383: Add published flag column to all user-writable tables
VERY IMPORTANT: add a spec that tests if there are no tables which could be populated by user, either via the submission system, or via API, and that do not have the new published flag.

Default all current data to published, not private.

*OR* add the flag field to *ALL* tables (which makes the spec mentioned above not required). This might be more practical option.

nowakowski:
Fixes #383

### DIFF
--- a/db/migrate/20160311121620_add_published_flag_to_db_columns.rb
+++ b/db/migrate/20160311121620_add_published_flag_to_db_columns.rb
@@ -1,7 +1,5 @@
 class AddPublishedFlagToDbColumns < ActiveRecord::Migration
   @@dbtables = [
-    :api_keys,
-    :countries,
     :design_factors,
     :genotype_matrices,
     :linkage_groups,
@@ -33,8 +31,7 @@ class AddPublishedFlagToDbColumns < ActiveRecord::Migration
     :taxonomy_terms,
     :trait_descriptors,
     :trait_grades,
-    :trait_scores,
-    :users
+    :trait_scores
   ]
 
   def up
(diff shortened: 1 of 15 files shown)


nowakowski:
Also fixes some old migrations to enable the test DB to be built from scratch.

Nuanda:
@kammerer Tomasz, perhaps you should do that review since you are going to use it.

db/migrate/20160311121620_add_published_flag_to_db_columns.rb
```diff
@@ -0,0 +1,50 @@
...
+    :qtl_jobs,
+    :restriction_enzymes,
+    :submission_uploads,
+    :submissions,
+    :taxonomy_terms,
+    :trait_descriptors,
+    :trait_grades,
+    :trait_scores,
+    :users
+  ]
```
kammerer: I am not convinced we need `:users`, `:api_keys` or `:countries` here.

kammerer:
:+1: Except for one minor comment.

Nuanda:
Correct, these three tables could be excluded from the list. :+1:

nowakowski:
Fixed. @kammerer - can I merge this now?

kammerer:
Sure.
12 mar 2016 16:26 "Piotr Nowakowski" <notifications@github.com> napisał(a):

Nuanda:
@nowakowski The schema file is wrong. You changed the migration but you did not rebuild the schema. Please do.

Issue #427, opened 2016-03-15
Extend PP submission 3rd step (new PL form) with sequence identifier


Nuanda:
Add the new field after plant variety but before annotations. The hint text should be as follows:

"Provide the identifier of the plant line sequence, in an external database, like SRA [link http://www.ncbi.nlm.nih.gov/sra/]. Identifiers starting with SR (like SRX1491141) will be automatically recognized as SRA submissions."

Issue #401, opened 2016-03-08
Adjust Browse and API sections to observe the new sequence id field


Nuanda:
PlantLines datatable should show the value of the identifier as an external link. Also, the API should mention that one is able to set that column value when submitting new data.

Handle possible submission adjustment as well.

Nuanda:
Link example: http://www.ncbi.nlm.nih.gov/sra/SRX1491141

Nuanda:
Also - test if ES indexed the seq.

PR #428, opened 2016-03-15
Plant line sequence identifier

Nuanda:
Fixes issues #400 and #401. Makes #427 possible.

Currently we only "recognize" SRA identifiers (i.e. we make them into hyperlinks in the data tables). The Ensembl links were, eventually, not discussed with Wiktor.


kammerer:
:+1:

PR #430, opened 2016-03-15
Submission visibility setting

### LINKED ISSUE #387: Setting submitted data as private or public
Outfit the final submission step (same for both current processes) with a setting that let's the user to set data as public or private. Make public the default.

The finalizer service should then take that flag into consideration for all created (but not referenced) objects.

kammerer:
Fixes #387.

Also, removes existing `#published?` method from models and marks related specs as pending. Maybe rename it to `#revocable?`?


Nuanda:
Let's wait with that a bit, so we have more information on how to proceed.

PR #431, opened 2016-03-16
Plant line sequence identifier submission

### LINKED ISSUE #427: Extend PP submission 3rd step (new PL form) with sequence identifier

kammerer:
Resolve #428 first.
Fixes #427.


Nuanda:
:+1:

PR #433, opened 2016-03-17
Fix published flag

### LINKED ISSUE #383: Add published flag column to all user-writable tables

nowakowski:
Fixes #383. As requested, I did not touch the existing implementation of `published?` in AR classes (requires further input from @wjurkowski).

### DIFF
--- a/app/services/submission/plant_population_finalizer.rb
+++ b/app/services/submission/plant_population_finalizer.rb
@@ -27,13 +27,8 @@ class Submission::PlantPopulationFinalizer
       attrs = attrs.with_indifferent_access
       taxonomy_term = TaxonomyTerm.find_by!(name: attrs.delete(:taxonomy_term))
       attrs = attrs.merge(
-        taxonomy_term_id: taxonomy_term.id,
-        entered_by_whom: submission.user.full_name,
-        date_entered: Date.today,
-        user: submission.user,
-        published: true,
-        published_on: Time.now
-      )
+        taxonomy_term_id: taxonomy_term.id
+      ).merge(common_data)
 
       if attrs[:plant_variety_name].present?
         plant_variety = PlantVariety.find_by!(plant_variety_name: attrs.delete(:plant_variety_name))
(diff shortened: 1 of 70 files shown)


Nuanda:
@kammerer Is that ok with you?

Nuanda:
Guys, let's wait with that one as well - I've just got more data from Wiktor. Will email you both.

nowakowski:
Keeping abreast of the dynamic discussion going on in #383 :). On @Nuanda's request I also dropped the `publishable` flag in `submissions`.

Nuanda:
Tested submissions manually and they seem to work ok. Please, take into consideration these three requests:
- [ ] separate adding timestamps to another migration
- [ ] the two PV->countries relations now have published without published_on - remove the published flag (you can still do that in the 20160317135309 migration and rename it afterwards)
- [ ] remove both new fields (published, published_on), and the validation, from all models that do not have user_id - I think it does not make sense to have e.g. private data with no ownership, right?

Nuanda:
Why did your migrations removed so many `on_update: :cascade` callbacks from schema.rb? Are they important or not?

nowakowski:
Hm; `schema.rb` is generated, not written. I'm not sure how exactly Rails generates this file but I've checked the model classes and there seems to be no reason to enforce cascading deletes on the DB level (there are no `dependent` conditions in AR classes). I believe the present form of `schema.rb` is correct.

app/services/submission/plant_population_finalizer.rb
```diff
@@ -30,7 +30,9 @@ def create_new_plant_lines
         taxonomy_term_id: taxonomy_term.id,
         entered_by_whom: submission.user.full_name,
         date_entered: Date.today,
-        user: submission.user
+        user: submission.user,
+        published: true,
+        published_on: Date.today
```
Nuanda: Here you use Date.today yet in other places you set that column to Time.now. Which is correct?

nowakowski: You're right, it should be `Time.now`. Fixed.

Nuanda:
@nowakowski Piotr, when you extend the correct code with new functions, please also extend corresponding specs do the new feature is being tested as well.

nowakowski:
@Nuanda Acknowledged. Is there anything specific to be done in the context of this PR?

Nuanda:
:+1:

Issue #404, opened 2016-03-08
Add 5th Browse section submenu item called Plant Trials


Nuanda:
It should show the Plant Trials table, the new pivot table (see #402) and, perhaps, some other tables (PSU? TS?). These tables should be moved from the other sections.

Issue #403, opened 2016-03-08
Add link to new 'pivot' Plant Trial table in the Plant Trial rows


Nuanda:
Each row in the original table should feature a link that leads the user to the new pivot table (#402).

Issue #402, opened 2016-03-08
Create data table for all "pivot" Plant Trial scores


Nuanda:
Shows all Plant Scoring Units (for a specific Plant Trial) with all their Trait Scores, in a single table.

Separate Plant Scoring Units should be represented by rows while related Trait Descriptors should be represented by columns, giving Trait Score values at the intersection of the appropriate row and column.

Make sure Export to CSV button works for this special table as well.

Nuanda:
See also #404

PR #436, opened 2016-03-18
Trial pivot table

Nuanda:
1. Adds a new Browse subsection (Trials) and moves 3 tables there; fixes issue #404.

2. Adds TrialScoring controller that creates the new "pivot" table for a PlantTrial; we (me and Wiktor) both know that for some trials is has to be much broader than the table width, but Wiktor accepts that; fixes issue #402.

3. Adds a link in PllantTrial table that shows the user the new TrialScoring table for a given PT; fixes issue #403.

### DIFF
--- a/app/assets/javascripts/data_grid.js.coffee
+++ b/app/assets/javascripts/data_grid.js.coffee
@@ -61,6 +61,8 @@ $ ->
           api = new $.fn.dataTable.Api(dt)
           if $(dt.nTHead).find('th.related').length
             api.columns().indexes().toArray().slice(0, -2)
+          else if dt.sInstance == 'trial-scoring'
+            api.columns().indexes().toArray()
           else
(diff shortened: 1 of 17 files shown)


kammerer:
@Nuanda Have you considered placing the new link in the "Related" dropdown?

kammerer:
:+1:

It probably could be done in a more uniform way (with regards to datatables-related code) by using database-view-backed readonly AR model, but this works fine too ;)

Nuanda:
1. I did consider placing it in the dropdown but this is a more than just a relation, so I decided to expose it a bit more. We'll see if Wiktor likes that.
2. I was planning to do that with no-DB model and even started that, but after a while I realized it is, in reality, much more work, and not as much code reusability as one would think. Hence this "special" controller.

Issue #434, opened 2016-03-18
Add short explanation what is returned in API create call


Nuanda:
And how to use it (mostly, the assigned id) to submit further data.

PR #437, opened 2016-03-18
Some info on create operation response

Nuanda:
Fixes issue #434

### DIFF
--- a/app/views/application/api.html.haml
+++ b/app/views/application/api.html.haml
@@ -251,9 +251,9 @@
         a correct value of <code>plant_population_id</code> will set the relationship.
 
       %p
-        If the submission finished with success, you will get a JSON document in return. It
+        If the submission is successful, you will get a JSON document in return. It
         describes the submitted record, so it mostly contains the data you have just submitted.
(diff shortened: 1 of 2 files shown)


app/views/application/api.html.haml
```diff
@@ -250,6 +250,19 @@
         resource (i.e., in a reverse fashion). That is, when submitting a new Plant Trial,
         a correct value of <code>plant_population_id</code> will set the relationship.
 
+      %p
+        If the submission finished with success, you will get a JSON document in return. It
```
nowakowski: Suggest: `If the submission is successful, (...)`.

app/views/application/api.html.haml
```diff
@@ -250,6 +250,19 @@
...
+        describes the submitted record, so it mostly contains the data you have just submitted.
+        However, it also provides the <i>internal BIP identifier</i> that was given to the new
```
nowakowski: `assigned` rather than `given`.

nowakowski:
Generally :+1: except for some minor changes in wording.

PR #451, opened 2016-03-25
Remove #published?, add #revocable?

### LINKED ISSUE #441: Rename and refactor #publishable? method
We need a new name for that, and we need it to work slightly differently. Its purpose is the following:
 - it says if the owner of data is able to set it from public to private
 - as a consequence, it says if the owner of data is able to API-delete it
 - it tells the other users if data is "quotable" (if it no longer can be made private/deleted) - this information is given inside the Metadata popup in the rightmost column in the datatables.

It should span 1 week from the published_on date.

kammerer:
Fixes #441


Nuanda:
:+1:

PR #460, opened 2016-04-01
Privacy aware browsing

nowakowski:
Implements privacy aware browsing in all model classes which require it.


nowakowski:
@kammerer - I need a bit of assistance with this one. I had to refactor the `Filterable` class since I need it to be able to operate on existing query objects instead of starting from scratch with `query=all`, and I also need it to return a proper `ActiveRecord::Relation` instead of an ElasticSearch result set (this is required for compatibility with the Squeel gem). Everything seems to work now but my changes broke a bunch of specs in `spec/models/concerns/filterable` - since we (@Nuanda and myself) weren't able to come up with a quick fix, I have commented out these tests and would like to ask for your assistance in repairing them - it appears that the problem is with the specs rather than with the code.

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,10 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  before_filter do |c|
+    User.current_user_id = User.find(c.session[:user]).id unless c.session[:user].nil?
+  end
```
Nuanda: @nowakowski Piotr, I believe the model.table_data method is called from a single place only, that is a single line inside the DataTablesController. That means you might be able to simply pass current_user as a parameter there, relieving User class from a surplus class variable and not making the above query, unnecessary, for any other non-data-table call. You will then be able to pass that current_user value down to 'visible' scopes, likewise. Wouldn't that be a better solution?

app/models/plant_accession.rb
```diff
@@ -12,13 +12,25 @@ class PlantAccession < ActiveRecord::Base
...
 
+  scope :visible, ->() {
+    uid = User.current_user_id
+    if uid.present?
+      where("published = 't' OR user_id = #{uid}")
+    else
+      where("published = 't'")
+    end
+  }
+
```
Nuanda: @nowakowski Piotr, remember to move it to a concern.

app/controllers/application_controller.rb
kammerer: Such way of sharing request state with model layer can lead to bugs as we use threaded application server. Simply passing user instance is the way to go.

kammerer:
@nowakowski Fixed failings specs. Before your changes they passed because `Search` returned a lazy result-set and now you asked it for ids which resulted in a call to ES (by default blocked in specs).

nowakowski:
@kammerer @Nuanda As requested, I got rid of the classvar in `User` and moved my scope to a concern. I also parameterized the `visible` scope with `user_id` due to the fact that model classes do not have access to the session and must obtain the ID of the current user from upstream (i.e. from the controller). Specs have been updated and I also merged the latest master back into my branch.

This is ready to be merged, IMO.

app/models/concerns/publishable.rb
```diff
@@ -6,6 +6,18 @@ module Publishable
 
     validates_with PublicationValidator
 
+    scope :visible, ->(uid = nil) {
+      if uid == nil
+        uid = Thread.current[:user]
+      end
```
Nuanda: @nowakowski Piotr, do we really need this? Perhaps it's ok to leave uid == nil?

PR #463, opened 2016-04-07
Zenodo deposition basis

### LINKED ISSUE #399: Create a service that submits the CSV data to Zenodo
It should return the registered Zenodo deposition DOI to the user.

### LINKED ISSUE #459: Set of data to be deposited to Zenodo for a given submission
For PlantPopulation submissions the following records are going to be uploaded to Zenodo:

- the submitted plant population
- female/male parent line, if present
- plant lines related to the plant population through the plant population list table
- plant varieties related to all exported plant lines

For PlantTrial submission:

- the submitted plant trial
- all (both new and old) trait descriptors used in the submission
- all plant scoring units and trait scores recorded during the submission
- all related plant accessions.

Currently the plan is to store them in separate CSV files (one per DB table) and send them in a single Zenodo deposition.

Is this data set correct?

Nuanda:
Fixes #399 fixes #459

You'll need a Zenodo account and an API key (see secrets) - see https://zenodo.org/dev for an instruction on how one gets a key.

TODOS:

- [x] new Submission methods (`object_name`, `object_description`, `affiliation`) should perhaps be moved to some kind of decorator
- [ ] input sanitization for contributors metadata entered by users
- [x] error management in Zenodo Depositor (test if users are informed what happened in case of BIP -> Zenodo communication error)
- [x] add `keywords` and `subjects` deposition metadata as described by Annemarie.


Nuanda:
@kammerer Tomasz, before we review this, you will need a Zenodo sandbox account and API key. Please visit https://sandbox.zenodo.org/, create an account, go to account/settings/applications/, generate a "Personal access token" and set it as `ZENODO_KEY` env variable in your `.env` file. Beware, Zenodo shows you this token _only once_!

app/controllers/depositions_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  before_filter :authenticate_user!
+
+  def new
+    @deposition = Deposition.new(submission: submission, user: current_user)
+  end
+
+  def create
+    @deposition = Deposition.new(
+      { submission: submission, user: current_user }.merge(deposition_create_params)
```
kammerer: Why not merge `submission` and `current_user` inside `deposition_create_params` method?

app/controllers/depositions_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+      render :new
+    end
+  end
+
+  private
+
+  def submission
+    if params[:deposition][:submission_id]
+      current_user.submissions.find(params[:deposition].delete(:submission_id))
+    else
```
kammerer: `else nil` is redundant.

Nuanda:
I hope the `submit_request`, `report_problem` and the `user_log` in the ZenodoDepositor service are enough for proper error management and reporting to the user.

app/models/submission.rb
```diff
@@ -99,8 +102,28 @@ def steps
...
+
+  def object_description
+    case submission_type
+    when 'population' then content.step01.try(:description) || ''
+    when 'trial' then content.step01.try(:plant_trial_description) || ''
+    else ''
+    end
+  end
+
+  def affiliation
```
kammerer: Why not `object_affiliation`? Or `name` instead of `object_name`, etc.

app/services/submission/exporter.rb
```diff
@@ -0,0 +1,31 @@
+class Submission::Exporter
+  def initialize(submission)
+    @submission = submission
+  end
+
+  def documents
+    raise NotImplemented, "Must be implemented by subclasses"
```
kammerer: It's `NotImplementedError`

app/helpers/application_helper.rb
```diff
@@ -70,6 +70,7 @@ def confirmable_action(label, object, options = {}, &blk)
     options[:method] ||= :post
     options[:url] ||= url_for(object)
     options[:btn_class] ||= "btn-default"
+    options[:button_size] ||= ""
```
kammerer: `:btn_size` (or `:button_class` and `:button_size`)?

spec/rails_helper.rb
```diff
@@ -100,3 +101,11 @@ def delete_document; end
 end
 
 OmniAuth.config.test_mode = true
+VCR.configure do |c|
+  c.cassette_library_dir = Rails.root.join('spec', 'support', 'vcr_cassettes')
```
kammerer: It seems natural to me to put cassettes dir into fixtures.

kammerer:
@Nuanda I looked at the code, so far I only found cosmetic issues.

Have you considered storing DOI in persistent `Deposition` model and relating it to `Submission`? Perhaps it's too far-fetched idea, but it would be easier to extend e.g. in case we needed multiple depositions per submission (or multiple submissions in one deposition).

app/models/submission.rb
Nuanda: This is actually not random, though it might look like it is. `object_` because it describe the object that was submitted, not the submission itself. Perhaps `submitted_object_name` would fit even better.

But `affiliation` is not something that describes the object. It is in fact the creator's affiliation. It is strange to think that a plant trial has an affiliation :). That's the whole reason.

spec/rails_helper.rb
Nuanda: Well, it's where I put them in the other project. So let's keep it that way.

Nuanda:
@kammerer Thanks for the review. Yes, I was considering that, but that would produce a single-column model, which is also referenced to from a single foreign table (thus, being a wrong design). So, I decided to stick with this "minimalistic" approach.

What do you think about the first TODO (see the OP)? Should we move these methods?

Nuanda:
(forgot to set the label)

Nuanda:
@kammerer Tomasz, any idea what should we do with the metadata input, especially with the contributors list, to make sure nothing is easily broken by strange/malicious input?

kammerer:
@Nuanda Since we are not displaying this metadata anywhere, we don't need to do anything - it's zenodo's responsibility to sanitize its input or otherwise handle it in a safe way. Having said that, we may try to verify if there is no malicious input (using Loofah) and abort deposition in such case (I don't think we should submit altered data to zenodo). But this is not going to be bullet-proof and may even give false-positives.

Nuanda:
I see. Apart from that, any other things I should consider before merging?

kammerer:
@Nuanda I think it can be merged as is. I am doing some refactorings but that better waits until your return from holidays.

Nuanda:
Should I delete the branch then, or do you want to keep that for your refactors?

kammerer:
I am using other branch, so I will delete this one.

PR #466, opened 2016-04-09
ES indexing takes into account record visibility

### LINKED ISSUE #389: Adjust ES indices not to include private data
This will probably result in private data owned by the logged in user also not searchable for her, but let's keep that simpler, and make only public data searchable.

Public data == data with no ownership OR data with published set to true.

kammerer:
Fixes #389


Nuanda:
+1

Nuanda:
:+1:

PR #467, opened 2016-04-12
453 published switch for submissions

### LINKED ISSUE #453: Add private/public switch to the submission list, for finalized submissions
Each finalized submission should have a visibility level shown in the submissions list (i.e. if its public or not). Add a switch to make private submissions public, and to make public submissions private (if possible).

When a submission was made public, add a short information somewhere by that switch, how many days left before data becomes "quotable" and the publication is no longer revocable. When past that time, disable the switch and explain that the data status cannot be made private any longer.

In case it is not clear what data (that is, what certain records in which tables) should be made private/public when this switch is used, report back to me - we'll find a viable solution.

kammerer:
Fixes #453 

@Nuanda There are no specs yet but the logic is more or less complete, so you can have a look.

One (possibly minor) issue is that at present there is no easy way to tell which submission resulted in creation of an associated object. As a result changing visibility of one submission may affect visibility of some objects associated with another submission belonging to the same user.

### DIFF
--- a/app/services/submission/publisher.rb
+++ b/app/services/submission/publisher.rb
@@ -9,6 +9,7 @@ class Submission::Publisher
 
   def publish
     raise ArgumentError, "Submission is published" if submission.publishable?
+
     now = Time.zone.now
     submission.transaction do
       submission.update_attributes!(publishable: true)
@@ -21,6 +22,8 @@ class Submission::Publisher
 
   def revoke
     raise ArgumentError, "Submission is not published" unless submission.publishable?
+    raise ArgumentError, "Submission is not revocable" unless submission.revocable?
+
     submission.transaction do
       submission.update_attributes!(publishable: false)
       revoke_object(submission.submitted_object)
--- a/app/services/submission/publisher.rb
+++ b/app/services/submission/publisher.rb
@@ -1,0 +1,46 @@
+class Submission::Publisher
+  attr_accessor :submission
+
+  def initialize(submission)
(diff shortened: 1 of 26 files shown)


Nuanda:
That's not a real issue. Anyway I'm aware of that and I don't think there is a way around that.

app/services/submission/publisher.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def publish
+    raise ArgumentError, "Submission is published" if submission.publishable?
+    now = Time.zone.now
+    submission.transaction do
+      submission.update_attributes!(publishable: true)
+      publish_object(submission.submitted_object, now)
+      associated_collections.each do |collection|
+        collection.not_published.each { |instance| publish_object(instance, now) }
```
Nuanda: There might be thousands of trait scores published with a submission. Having them published one by one may take some time. Would it be possible to optimize it a little with with update_all?

Nuanda:
Had a code scan and yes, this is what is needed :). Please go ahead with specs. I made one remark re. potential optimization.

Nuanda:
Also tested on a PP submission and it looks ok. One additional idea - please repeat the publish/revoke buttons on Submission#show view as well.

kammerer:
Updated.

I did not include any publishing optimizations for now since it is not that straightforward. But if there are thousands of associated objects then we'll probably need to make these operations async at the very least.

Nuanda:
:+1:

Issue #390, opened 2016-03-07
Tune cache mechanism to still work properly for public/private data


Nuanda:
See also #384 

We want that browse section tables use cache for public access (no user logged in), but respect private data for user logged in.

Please investigate and decide what caching strategy would be suitable here - please post proposed solution before implementing.

Add specs to make sure we do not show (the non-owner user) data, that has been made private, just because it was published before then (i.e. when the cache was built).

PR #468, opened 2016-04-12
Makes "Related" counters updated by number of invisible records

### LINKED ISSUE #456: Manage counters of related models to include/exclude private relations

Nuanda:
Fixes #456 

The visibility of the related records depends on the usual algorithm. Counters should be updated correctly.

### DIFF
--- a/app/models/concerns/pluckable.rb
+++ b/app/models/concerns/pluckable.rb
@@ -13,10 +13,16 @@ module Pluckable extend ActiveSupport::Concern
   included do
     include Joinable
 
-    def self.pluck_columns
+    def self.pluck_columns(uid = nil)
       query = self.all
-      cc = respond_to?(:count_columns) ? count_columns : []
-      columns = table_columns + cc + ref_columns
+      # Detect Relatable models
+      count_columns = if self.respond_to?(:privacy_adjusted_count_columns)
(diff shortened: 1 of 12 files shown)


Nuanda:
One quite minor thing is with PP - PPL - PL relation. If a PP has a public PPL, that has a private PL, the counter on PP will still show 1 (though navigating there will result in an empty table, so no privacy break there). I'm afraid fixing that single case would be quite difficult so I do not intend to do that, unless some simple fix is found (or more cases like that will appear after model changes).

Also, please remember to deactivate DTController cache before testing, as #390 is still open.

kammerer:
:+1: 

Seems to work OK. Not the easiest thing to read, not that I have any idea how to improve it.

As for PP - PPL - PL relation, I think we can add a callback to PL changing PPL visibility according to PL's visibility. It should be safe to do that both ways.

Nuanda:
Did you mean something like that?

kammerer:
Yup

2016-04-18 23:14 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Did you mean something like that?
>
> —
> You are receiving this because you were assigned.
> <https://github.com/eSpectrum-IT/brassica/pull/468#issuecomment-211581982>
>



-- 
Tomasz Szymczyszyn

PR #474, opened 2016-04-14
Invalidation of cache on different user and when related records are modified

### LINKED ISSUE #390: Tune cache mechanism to still work properly for public/private data

Nuanda:
Fixes #390 

It's still not perfect but better than it was. It's not only about privacy but rather about relations. Now I try to invalidate cache:
 - when a different user sees the data (per-user-cache)
 - when Related objects change (disappeared, were hidden)
 - when joined objects that have columns listed in the relating object view, change (or a re removed).

Sometimes we have joins traversing several models and this is not supported (I don't believe 'touch' executes callbacks and it means no cascading touches).

'before_destroy' is used since otherwise fkeys are already nullified, and it also needs to be before has_many (which, I believe, adds its default :nullify as another 'before_destroy' callback).

I provide only a single case testing as all-cases generic testing is rather tricky to write and would make specs dragging too much.

At the end of the day, if time and budget permits, we'll probably go for server-side, paginated responses, which will make tables caching not that important.

### DIFF
--- a/app/models/plant_population.rb
+++ b/app/models/plant_population.rb
@@ -7,26 +7,22 @@ class PlantPopulation < ActiveRecord::Base
              foreign_key: 'female_parent_line_id'
   belongs_to :user
 
-  has_many :linkage_maps,
-           dependent: :nullify
-  has_many :population_loci,
-           dependent: :nullify
-  has_many :plant_trials,
-           dependent: :nullify
-
-  has_many :plant_population_lists, dependent: :delete_all
-  has_many :plant_lines,
-           through: :plant_population_lists
-
-  validates :name,
-            presence: true,
-            uniqueness: true
-  validates :user,
-            presence: { on: :create }
-
   after_update { population_loci.each(&:touch) }
   after_update { linkage_maps.each(&:touch) }
   after_update { plant_trials.each(&:touch) }
+  before_destroy { population_loci.each(&:touch) }
+  before_destroy { linkage_maps.each(&:touch) }
+  before_destroy { plant_trials.each(&:touch) }
+
+  has_many :linkage_maps
+  has_many :population_loci
+  has_many :plant_trials
+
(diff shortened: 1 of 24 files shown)


kammerer:
:+1: 

1. I merged in master and added a few specs.
1. Can you give an example of a model which needs its cache invalidated when model related in two steps is updated? 
1. `touch` only triggers `after_touch` (and `after_commit/rollback`)

kammerer:
Oh, it looks strange to me that you put callbacks before `has_*` relations in models.

Nuanda:
1. Thanks.
2. Not many of them. TraitScores show a lot of data in their table - for instance, PP.name - if one changes the PP in question visibility to private, TraitScores table will still show its name.
3. Is this good or bad? :)
4. I explained that in the PR description above - when declared after cascade, it is going to be executed after rails set fkeys to nil, hence it won't have any effect.

app/models/plant_population.rb
```diff
@@ -7,6 +7,13 @@ class PlantPopulation < ActiveRecord::Base
              foreign_key: 'female_parent_line_id'
   belongs_to :user
 
+  after_update { population_loci.each(&:touch) }
+  after_update { linkage_maps.each(&:touch) }
+  after_update { plant_trials.each(&:touch) }
+  before_destroy { population_loci.each(&:touch) }
+  before_destroy { linkage_maps.each(&:touch) }
+  before_destroy { plant_trials.each(&:touch) }
+
```
Nuanda: Thanks. Forgot about this one.

Nuanda:
What should we do about that one?

PR #475, opened 2016-04-15
Block users getting other's private data

Nuanda:
This also tests deleting not owned data and setting published and published_on flags on create (untested before).

### DIFF
--- a/app/controllers/api/v1/resources_controller.rb
+++ b/app/controllers/api/v1/resources_controller.rb
@@ -27,11 +27,16 @@ class Api::V1::ResourcesController < Api::BaseController
   end
 
   def show
-    resource = model.klass.where(id: params[:id])
-    resource = resource.visible(api_key.user_id) if resource.respond_to? :visible
-    resource = resource.first
-    response_json = resource ? { model.name => decorate(resource) } : {}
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/resources_controller.rb
```diff
@@ -19,25 +19,29 @@ def index
...
     render json: { model.name.pluralize => resources, :meta => resources.meta }
   end
 
   def show
-    resource = model.klass.find(params[:id])
-    render json: { model.name => decorate(resource) }
+    resource = model.klass.where(id: params[:id])
+    resource = resource.visible(api_key.user_id) if resource.respond_to? :visible
+    resource = resource.first
+    response_json = resource ? { model.name => decorate(resource) } : {}
```
kammerer: Shouldn't we return 403 or 401 instead of empty response? We do that in `destroy` action.

kammerer:
:+1: 

One minor concern with returned response for hidden resource.

app/controllers/api/v1/resources_controller.rb
Nuanda: Ok.

PR #476, opened 2016-04-15
Prevents user from seeing another person private data

### LINKED ISSUE #461: Submission autocompletes should not propose private, not owned data
For instance, the 2nd step of the Plant Trial submission should propose either public TDs, or TDs that belong to the current_user.

Wait for appropriate 'visible' scopes to arrive to the models, before implementing this.

Nuanda:
Fixes #461 

Please note this is not to prevent the user relate her records to someone else's hidden records - this is, actually, not forbidden (one may do that by guessing an id). I just wanted to prevent users seeing (in the autocomplete list) information about other user's hidden records.

### DIFF
--- a/app/controllers/plant_lines_controller.rb
+++ b/app/controllers/plant_lines_controller.rb
@@ -2,7 +2,7 @@ class PlantLinesController < ApplicationController
 
   def index
     page = params[:page] || 1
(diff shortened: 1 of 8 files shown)


kammerer:
:+1:

PR #479, opened 2016-04-18
Faster inter-submission navigation

### LINKED ISSUE #409: Faster inter-submission process navigation
Make the circular "metro line station" markers clickable so a user may quickly go back >1 step in the submission process. She should also be able to go forth >1 step as long as no unsubmitted step is overlooked.

kammerer:
Fixes #409 

Please, have a look and tell me if it works for you. Tested in FF/linux.
 
- You should be allowed to click on step numbers for steps with some content already present
- You should be asked for confirmation if you make some changes and click on step number or go back
- You should be asked for confirmation if you make some changes to new plant line / trait descriptor and try to go forward (but there should be no confirmation if you confirm or cancel it)


Nuanda:
In my opinion there is no need for a better confirmation, standard browser dialog box is enough here. One small UX enhancement would be: if you are able to detect that a user is on an earlier step (1), and is trying to "go forth" to a later step (2), the confirmation dialog may suggest that the correct action to _save_ the changes is to click NEXT or SAVE & EDIT LATER.

The latter button (SAVE & EDIT LATER) may be suggested when one is trying to "go back" to a previous step, with the upper "metro line" navigation, while having some unsaved changes.

Otherwise :+1:

kammerer:
I am not sure about the default confirmation dialog. It looks OK in Firefox but IMHO not so much in Chrome.

I changed the confirmation messages shown when user tries to advance submission as you suggested. I am not sure about the message when he goes back, it does not seem sensible to me to suggest clicking SAVE & EDIT LATER, as it takes user to the index view. It would be sensible to use SAVE NOW, but we don't have such a button. Maybe we should change SAVE... button into a dropdown button with two options? Not sure if that's not an overkill.

Nuanda:
I think it's fine the way it is now. :+1:

kammerer:
Added model and request specs. Fixed some javascript issues. Tested in FF/linux, Chromium/linux, IE10 and 11. I think it works reasonably well. There is IMHO minor issue with fields which allow to add new or select existing TDs / PLs:
- if there is an existing item in the list when you visit relevant step 
- you remove this item and then add it again
- tracking mechanism still sees that a change was made

I think it can be ignored. Ready for review.

Nuanda:
:+1:

PR #480, opened 2016-04-20
Publish state machine fix

Nuanda:
One fix is the state machine - I realized one is able to "Publish" already published, and no longer revocable, submission. I also moved both buttons to the _publishing partial.

Another "fix" is strictly visual: when I clicked "no" when asked to confirm "Publish" (or Revoke or Delete, for that matter), it triggered an ugly collapse "animation". Removing the upper margin from the .actions div in the table made the ugly effect disappear. However, now we have the buttons misaligned (changing margin-top to padding-top made no difference). Another solution - http://stackoverflow.com/questions/13119906/turning-off-twitter-bootstrap-navbar-transition-animation - works by making the transition "ultra-fast" (but sometimes still not fast enough). If you have other ideas how to solve that, please commit them here.

### DIFF
--- a/app/views/submissions/_publishing.html.haml
+++ b/app/views/submissions/_publishing.html.haml
@@ -1,0 +1,6 @@
+- if submission.publishable? && submission.revocable?
+  = confirmable_action("Revoke publication", submission,
+      url: revoke_submission_path(submission), method: :patch)
+- elsif !submission.publishable?
+  = confirmable_action("Publish", submission,
+      url: publish_submission_path(submission), method: :patch, btn_class: "btn-primary")
--- a/app/assets/stylesheets/partials/_submissions.scss
+++ b/app/assets/stylesheets/partials/_submissions.scss
@@ -19,7 +19,6 @@
     .actions {
       float: right;
       width: 17%;
-      margin-top: 6px;
       text-align: right;
     }
   }
--- a/app/views/submissions/_submissions_list.html.haml
+++ b/app/views/submissions/_submissions_list.html.haml
@@ -30,10 +30,4 @@
               class: 'btn btn-sm btn-default'
 
         - else
-          - if submission.publishable? && submission.revocable?
-            = confirmable_action("Revoke publication", submission,
-                url: revoke_submission_path(submission), method: :patch)
-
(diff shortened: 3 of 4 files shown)


app/views/submissions/_publishing.html.haml
```diff
@@ -0,0 +1,6 @@
+- if submission.publishable? && submission.revocable?
+  = confirmable_action("Revoke publication", submission,
+      url: revoke_submission_path(submission), method: :patch)
+- elsif !submission.publishable?
```
kammerer: I still think that it should be `Submission#published?` and not `#publishable?`. Now it looks like the "Publish" button is shown for submissions which are not publishable :)

kammerer:
As for the visual glitch - I think the best way to fix it is to write a behaviour similar to collapse but not relying on height change. However, I am going to introduce a confirmation modal in #479, so perhaps we should use it here as well?

kammerer:
:+1: for state machine fix and partial extraction.

app/views/submissions/_publishing.html.haml
Nuanda: What can I say - I though this was agreed upon between you and @nowakowski - but perhaps I've got that wrong. Do you think we should change Submission.publishable to Submission.published?

Nuanda:
I precisely used that Yes/No confirmation as I think it is much better compared to the modal. Modal is required in case you need to explain something to the user: what would be the consequence of this operation.

So - yes and no :). I would stick with Yes/No confirmation with the Delete operation, but we might use modals for Publish/Revoke, to explain further what is going to happen (and what is that 'grace period'). Otherwise, we may put that explanation in a tooltip. I leave it to you.

So, should I revert that css change?

app/views/submissions/_publishing.html.haml
kammerer: I got the impression that it was agreed between you and @nowakowski :) I believe that `#published?` makes more sense because it de facto indicates that submission's data is published and not that it can be published (without saying anything about current state).

kammerer:
You can leave it as is. I will make another issue to remind about this.

PR #484, opened 2016-04-23
Publish/revoke API

### LINKED ISSUE #388: Extend API with published flag
Since the DB tables and the model are going to be extended with the published flag, the API should automatically both show and set that flag, when querying/creating data. Just make sure it works that way.

Extend the API docs by the description of that concept.

kammerer:
Fixes #388 

I extended the docs but the wording could be improved.

### DIFF
--- a/app/models/concerns/publishable.rb
+++ b/app/models/concerns/publishable.rb
@@ -21,7 +21,7 @@ module Publishable
     }
 
     before_validation -> {
-      self.published_on ||= Time.zone.now if published?
+      self.published_on ||= Time.now if published?
     }
 
     def revocable?
--- a/app/models/concerns/publishable.rb
+++ b/app/models/concerns/publishable.rb
@@ -31,5 +31,16 @@ module Publishable
     def private?
       !(published? || user.nil?)
     end
+
+    def publish
+      return if published?
+      update_attributes!(published: true, published_on: Time.zone.now)
+    end
+
+    def revoke
+      return unless published?
+      raise "#{self.class}##{id} cannot be revoked" unless revocable?
+      update_attributes!(published: false, published_on: nil)
+    end
   end
 end
--- a/app/models/concerns/publishable.rb
+++ b/app/models/concerns/publishable.rb
@@ -20,6 +20,10 @@ module Publishable
       where condition
     }
 
+    before_validation -> {
+      self.published_on ||= Time.zone.now if published?
+    }
+
     def revocable?
       published? && Time.now < revocable_until
     end
@@ -34,7 +38,7 @@ module Publishable
 
     def publish
       return if published?
-      update_attributes!(published: true, published_on: Time.zone.now)
+      update_attributes!(published: true)
     end
 
     def revoke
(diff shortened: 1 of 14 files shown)


Nuanda:
I think the `published: false` functionality for resource creation is not yet enabled, right? The API manual says it is implemented, so I suggest we add that in this PR already, ok?

kammerer:
Hmm, I forgot to add spec for that, but I think it should work. I will update the PR with such spec.

kammerer:
Updated PR. It was not only the spec after all.

app/models/concerns/publishable.rb
```diff
@@ -20,6 +20,10 @@ module Publishable
       where condition
     }
 
+    before_validation -> {
+      self.published_on ||= Time.zone.now if published?
```
Nuanda: Why Time.zone.now instead of Time.now? If it writes to DB I think one is safe to rely on Time.now to return the UTC time (which we use internally). Isn't Time.zone.now only for presentation?

app/services/api.rb
```diff
@@ -70,4 +74,9 @@ def self.writable_model?(model)
     @writable ||= writable_models.map { |m| m.name.underscore }
     @writable.include?(model.to_s.underscore.singularize)
   end
+
+  def self.publishable_model?(model)
```
Nuanda: Is this only used in specs?

app/models/concerns/publishable.rb
kammerer: It is better to preserve timezone information when writing to DB, but I just checked the tables and it appears that "timestamp without time zone" is used for datetime columns, so it should not matter if we use `Time.now` or `Time.zone.now`.

app/services/api.rb
kammerer: Yup

Nuanda: I'd rather not have application code that is only used in specs - I try to avoid it. If possible, please move that method to the corresponding spec file.

app/models/concerns/publishable.rb
Nuanda: Actually we rather try to use this technique in our project - i.e. saving all timestamps in UTC only in the DB. So please change to Time.now to remain consistent.

app/services/api.rb
kammerer: I forgot it is also used for setting constraints on routes so I am leaving it as is.

app/models/concerns/publishable.rb
kammerer: Changed to `Time.now` to avoid confusion.

BTW It returns time in system TZ, not UTC time (but here it does not matter anyway).

app/services/api.rb
Nuanda: But you use only the 'publishable_models' there, right?

kammerer: Right, removed `Api.publishable_model?`.

kammerer:
@Nuanda Everything you mentioned so far is fixed.

Nuanda:
Thanks! :+1:

PR #489, opened 2016-04-25
TSV to CSV for trait scoring

### LINKED ISSUE #442: Switch user data upload from TSV to CSV
A proposition. We use CSV format for data download, yet we happened to set TSV format for data upload (in the Plant Trial submission). Perhaps we should switch to CSV data upload?

One thing I dislike about TSVs is that many editors deliberately change tabs to spaces (tabs in source code is the root of all evil, as we well know ;)) - no such problem with CSVs. While we (and users) use dots as fraction separators, there should be no problem with numerical data in CSVs, right?

@wjurkowski @teatree1212 what do you think?

Nuanda:
Fixes #442

### DIFF
--- a/app/services/submission/trait_score_parser.rb
+++ b/app/services/submission/trait_score_parser.rb
@@ -76,10 +76,10 @@ class Submission::TraitScoreParser
   end
 
   def csv
-    @csv || (@csv = CSV.new(input))
+    @csv ||= CSV.new(input)
   end
 
   def input
-    @input || (@input = File.open(@upload.file.path))
+    @input ||= File.open(@upload.file.path)
   end
 end
--- a/app/services/submission/trait_score_parser.rb
+++ b/app/services/submission/trait_score_parser.rb
@@ -34,16 +34,20 @@ class Submission::TraitScoreParser
 
   def map_headers_to_traits
     @upload.log "Mapping file header columns to Trait Descriptors"
-    header = input.readline
+    header = csv.readline
     @trait_names = PlantTrialSubmissionDecorator.decorate(@upload.submission).sorted_trait_names
     @trait_mapping = {}
-    header.split("\t")[1..-1].each_with_index do |column_name, i|
-      next if i >= @trait_names.length
-      column_name = column_name.strip
-      trait_index = @trait_names.find_index(column_name)
-      trait_index = i unless trait_index.present?
-      @trait_mapping[i] = trait_index
-      @upload.log " - Mapped column '#{column_name}' to Trait index #{trait_index} of value #{@trait_names[trait_index]}"
+    if header.blank?
+      @upload.errors.add(:file, 'No header provided.')
+    else
+      header[1..-1].each_with_index do |column_name, i|
+        next if i >= @trait_names.length
+        column_name = column_name.strip
+        trait_index = @trait_names.find_index(column_name)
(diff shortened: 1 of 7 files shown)


app/services/submission/trait_score_parser.rb
```diff
@@ -34,16 +34,20 @@ def call
...
     @trait_names = PlantTrialSubmissionDecorator.decorate(@upload.submission).sorted_trait_names
     @trait_mapping = {}
-    header.split("\t")[1..-1].each_with_index do |column_name, i|
-      next if i >= @trait_names.length
-      column_name = column_name.strip
-      trait_index = @trait_names.find_index(column_name)
-      trait_index = i unless trait_index.present?
-      @trait_mapping[i] = trait_index
-      @upload.log " - Mapped column '#{column_name}' to Trait index #{trait_index} of value #{@trait_names[trait_index]}"
+    if header.blank?
```
kammerer: Is this even possible? I think only when an empty file is uploaded, but then the "Choose file..." button remains disabled and nothing happens, as paperclip detects content-type spoofing.

`[paperclip] Content Type Spoof: Filename empty.csv (["text/comma-separated-values", "text/csv"]), content type discovered from file command: application/octet-stream. See documentation to allow this combination.
`

I think we can check if the first line begins with 'Plant scoring unit name' and assume the file is not formatted correctly otherwise.

app/services/submission/trait_score_parser.rb
```diff
@@ -71,6 +75,10 @@ def parse_scores
     @upload.log "Parsed #{score_count} scores for #{plant_count} plants, in total."
   end
 
+  def csv
+    @csv || (@csv = CSV.new(input))
```
kammerer: Don't we usually write it like this: `@csv ||= CSV.new(input)`? Nothing wrong with it, just surprised me a bit.

app/services/submission/trait_score_parser.rb
Nuanda: You are right but that's exactly why I did this. Paperclip's content spoofing detection is so faulty, I was recently forced to disable it in the other project (it completely confused one file type with another). I have a feeling that sooner or later we'll disable it here as well :(. Then, this will help us.

First line is not required to start with 'Plant scoring unit name' - our template suggests it but one is able to use own's template (and then the first column header is simply ignored).

app/services/submission/trait_score_parser.rb
Nuanda: Changed.

app/services/submission/trait_score_parser.rb
kammerer: OK, I think it's fine if you disable it.

Maybe it should be required to start with 'Plant scoring unit name'? It would help to recognize if there is a header line at all. Right now if you skip header altogether the data row (the first row) will be interpreted as header. Of course, that is user's fault, but perhaps we could handle that in a better way.

Nuanda: Maybe - but this piece of submission is soon going to be considerably rebuilt anyway. I'm in the process of discussing with @teatree1212 and @wjurkowski how we are going to extend/change this. So it's not that important right now. I simply wanted to switch from TSV to CSV _before_ we start to add/change upload columns (less work with format change in the future).

kammerer:
:+1:

PR #496, opened 2016-05-03
Fixed margins for action buttons in submissions table

### LINKED ISSUE #482: Fix margins for action buttons in submissions table
See: https://github.com/eSpectrum-IT/brassica/pull/480#issue-149869481

kammerer:
Fixes #482

### DIFF
--- a/app/assets/javascripts/lib/replace.js
+++ b/app/assets/javascripts/lib/replace.js
@@ -1,0 +1,188 @@
++function($) {
+  'use strict';
+
+  // REPLACE PUBLIC CLASS DEFINITION
+  // ================================
+
+  var Replace = function (element, options) {
+    this.$element      = $(element)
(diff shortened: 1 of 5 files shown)


Nuanda:
Confirmed in FF and Chromium. Can you check it in the IE?

Also, what does the new replace library doing? Can I use it to redefine the style for an 'externally defined' css class?

kammerer:
I will check IE. 

Replace works just like Bootstrap's Collapse but it does not modify dimensions, so right now it just changes display property of affected elements.

Nuanda:
Ah, I didn't scroll it down till the end - I see it's just a different class there.

Nuanda:
It's :+1: if it works in IE.

kammerer:
Works in IE11 at least.

PR #504, opened 2016-05-16
Add technical_replicate_number to trait_scores

### LINKED ISSUE #488: minimum requirements trial submission
Based on a meeting today, I have the minimum requirements which should be made compulsory in the trial submission: Year, (besides the country, also the) Place name, units, Block/pot number, replicate , line/cultivar/variety + , its genetic status. Optional internal accession fields would be desirable ( which the public doesn't necessarily need to see, but are important for internal analysis).
Further, the state the data is in should be specified. maybe you could make 3 tick boxes for 1) arithmetric mean- 2)harmonic mean- 3)raw data.

BUT: All these should only be made compulsory, if the data is raw data. If it is submitted based on a mean, then block/pot number and replicate are unnecessary.

Also, the Plot overview would be desirable. I think though, that this may need to be included in an additional development step (?), have a think: 
I have attached an image on what it looks like. However, they are always part of the excel spreadsheet they use for the field trial setup ( see screenshot). 
screen shot 2016-04-25 at 16 54 15

nowakowski:
Fixes #488

### DIFF
--- a/db/migrate/20160516160911_add_technical_replicate_number_to_trait_scores.rb
+++ b/db/migrate/20160516160911_add_technical_replicate_number_to_trait_scores.rb
@@ -1,0 +1,5 @@
+class AddTechnicalReplicateNumberToTraitScores < ActiveRecord::Migration
+  def change
+    add_column :trait_scores, :technical_replicate_number, :integer, default: 1, null: false
+  end
+end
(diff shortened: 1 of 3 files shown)


Nuanda:
Wrong schema.rb? Why is it changed so much?

nowakowski:
Hm, very odd... I rebuilt the DB from scratch for testing - I don't know why the `published` indices were missing in the final product; they should have been duly added by `20160415135850_add_missing_indices.rb`. Reverting this migration and running it again fixed the issue.

Nuanda:
:+1:

PR #513, opened 2016-05-27
Rename Submission#publishable to #published

### LINKED ISSUE #483: Change submission publishable flag name to published
See #480 discussion. I think I insisted on a different name for submissions model, but now I see 'publishable' is not good either. Perhaps we should revert it to 'submission#published'.

kammerer:
Fixes #483 

Also, renames "publishability" to "visibility".


Nuanda:
:+1:

Issue #423, opened 2016-03-08
Links to Planteome Trait Ontology in Browse section tables


Nuanda:
Probably only for the Trait Descriptor table (the other tables should rather link to that BIP table first, before going outside) - the Trait term should link to Planteome, if possible.

Link example: http://browser.planteome.org/amigo/term/TO:0000387

Issue #415, opened 2016-03-08
Links to Planteome Plant Ontology in Browse section tables


Nuanda:
When there is a plant part column in a datatable, it should be an external link to planteome term, if the given plant part record has a plant ontology term id assigned.

Link example: http://browser.planteome.org/amigo/term/PO:0000033

Issue #421, opened 2016-03-08
Import curated Trait Ontology to BIP


Nuanda:
Use a single-time rake task to import curated TO file to BIP data table - see #419 and #420 for information on where to import that.

Nuanda:
The `Trait` table is ready to accept more trait definitions. We may now use the google doc provided by @teatree1212 to load further terms (which are not present in TO).

Nuanda:
Since I have loaded all Trait Ontology terms already, and then added all BIP terms, only non-TO terms provided in #420 are left (usually curated by Annemarie from projects' lists). I will load them to the Trait table.

PR #516, opened 2016-05-28
Rebuilt TD model and 2nd PT submission step

### LINKED ISSUE #414: Import Plant Ontology to Plant Part table
Use a one-time rake task for that. PO will be provided by #412. Don't forget to store the ontology term id in the new column added in #413

### LINKED ISSUE #422: Adjust Plant Trial submission to new Trait Ontology functionality
Depending on #419, the new Trait Description wizard may need a selector to set Trait name as a relation to the new Trait "dictionary" table.

If not, we may need to provide an additional ontology_term_id field there, so the user may set Planteome term id by herself, when creating a new Trait Descriptor.

### LINKED ISSUE #505: Rebuild TraitDescriptors browse section table
Agreed with @wjurkowski (via an email discussion) that the current, convoluted TD table should be replaced by a more plain Traits list.

Some required fields:
  - keep the Trait scores link (might be moved to the Related dropdown)
  - add Trials link to the Related dropdown as well
  - remove all project-related columns (Species, Population, Project, Country)
  - add have more trait-related columns. Definitely both 'Units of measurement', 'Scoring method' and 'Materials' columns (see #503)
  - also, if we solve #416, perhaps 'Plant part' as well.

### LINKED ISSUE #507: Move PlantPart relation from PSU to TD
As decided in #416 please:
 - add plant_parts relation to trait_descriptors table
 - populate it with the plant part that is referenced by PSUs related to a given TraitDescriptor (should be consistent, all PSUs of a TD should have the same plant part); sometimes its nil or strange values (see #412); you get related PSUs through the TraitScores model
 - remove the PSU plant_part relation when done.

### LINKED ISSUE #508: Extend new TD form (Trial submission process) for the PlantPart relation
Wait for #507 first.

Add a Plant Part field where people can pick a PlantPart to be related with the Descriptor. In a similar fashion to the TaxonomyTerm select, in Population submission step 2. It should not be required.

### LINKED ISSUE #514: Change order of app::bootstrap ES index and migrations steps
Otherwise, some new columns do not get indexed in the ES.

Nuanda:
This is big - I wanted to put many closely-related changes in a single PR. In short, I did the following:
 - parsed PO ontology to (extended) PlantPart table and TO ontology to (new) Trait table
 - moved PP relation from PSU level to TD level, and related TD with the new Trait "vocabulary" table
 - created "non-canonical" PP and T entries to accommodate plant parts that are not inside PO and TD.descriptor_names that are not in TO (followed TaxonomyTerm table convention)
 - rebuilt the second step of the PlantTrial submission:
   - shorter field list as requested by Annemarie
   - uses T.name instead of TD.descriptor_name now
   - let's the user to pick a PlantPart
   - also, validations and `null: false` constraints are changed due to a different set of required TD columns requested by Annemarie
 - rebuilt the TD browse section data table, as discussed with Wiktor:
   - less convoluted joins to have a better "trait list" overview
   - rather trait-related than project-related fields in the table
   - the model was refactored accordingly.

Important things to note:
 - we should no longer use TD.descriptor_name column for any purpose, but it is left in the DB for curation
 - I did use rake tasks in migrations, as I don't see a better way to keep the entire migration process consistent (btw I changed places of the "migration"/"elastic indexes" steps in the app bootstrap rake task).

Fixes #514 fixes #508 fixes #507 fixes #505 fixes #422 fixes #414 

Partially related to #488 and to #421 

@nowakowski Piotr: I would ask you to review the migrations and the model changes. Please note one migration irreversibly drops a column, so please check if the curation logic is proper, as otherwise we risk data restoration on the production.

@kammerer Tomasz, please review the submission and finalizer changes; also see if ES indexing for the changed TD model works ok. See the email I sent you about the two lines in the form02 view (I left them commented out in this PR your you to use, if they are needed after all). Also, if you can (easily) change the PTFinalizer behavior from looking up Ts and PPs by names to looking them up by ids, please do that (I was not able to get it). Also, please check the API changes.


Nuanda:
This is RFR - there is a strange spec fail in a service I never changed, and it does work on my local setup  - so I guess it is safe to review it now.

app/decorators/plant_trial_submission_decorator.rb
```diff
@@ -45,7 +45,7 @@ def trait_descriptors
     tds = object.content.step02.trait_descriptor_list.try(:select, &:present?) || []
     numeric_tds = tds.select { |td| td.to_s.match /\A\d+\z/ }
     non_numeric_tds = tds - numeric_tds
-    non_numeric_tds | TraitDescriptor.where(id: numeric_tds).pluck(:descriptor_name)
+    non_numeric_tds | TraitDescriptor.includes(:trait).references(:trait).where(id: numeric_tds).select('traits.name')
```
kammerer: The name is somewhat misleading, but this method should return TD names, not records, so `pluck` not `select` should be used. Probably it would be good to change method name too.

Nuanda: Changed. Also added a spec to test for future regression like that.

Nuanda:
I think I was also able to fix the transient problem with indices on published and user_id columns, that was troubling @nowakowski recently.

app/views/submissions/steps/trial/_step02.html.haml
```diff
@@ -1,17 +1,17 @@
...
-  Examples of plant traits in the database are: 'oleic acid content' or ' seed mature time'
+  Examples of plant traits in the database are: 'oleic acid content' or ' seed mature time'.
 
 .form-group
   = content.label :trait_descriptor_list
-  - options = options_for_submission_select(@content.existing_trait_descriptors, :descriptor_name)
-  - options.concat options_for_submission_select(@content.new_trait_descriptors, :descriptor_name, id: false)
-  = content.select :trait_descriptor_list, options.html_safe, {},
+  -#- options = options_for_submission_select(@content.existing_trait_descriptors, :descriptor_name)
+  -#- options.concat options_for_submission_select(@content.new_trait_descriptors, :descriptor_name, id: false)
```
kammerer: These two lines are needed to show selected values when user returns to this step. Uncommenting fixes showing existing TDs, however it seems there is still an issue with new TDs (e.g. new TD is shown if there is also existing TD selected, but not if there is only new TD selected).

Nuanda: I see, thanks for explaining that to me. The issue might be due to the change in the following line - it does not use the `options` variable anymore. Can you test after reverting the change in the following line?

kammerer: Of course I reverted [that line](https://github.com/eSpectrum-IT/brassica/pull/516/files#diff-21107b65d30dbf80989b502ee8d40a70R14) too. Otherwise uncommenting would have no effect at all.

config/initializers/paperclip.rb
```diff
@@ -1,2 +1,11 @@
 Paperclip::Attachment.default_options[:url] = "/system/:class/:hash.:extension"
 Paperclip::Attachment.default_options[:hash_secret] =  Rails.application.secrets.paperclip_hash_secret
+
+require 'paperclip/media_type_spoof_detector'
+module Paperclip
+  class MediaTypeSpoofDetector
+    def spoofed?
+      false
```
kammerer: A pity spoofing detection cannot be disabled, it seems it's one of the most buggy features of paperclip.

config/locales/api.en.yml
```diff
@@ -434,9 +441,9 @@ en:
         -
           name: confirmed_by_whom
         -
-          name: plant_scoring_unit_ids
+          name: trait_descriptors_ids
```
kammerer: It should be `trait_descriptor_ids`.

config/initializers/paperclip.rb
Nuanda: :) Actually this commit is not needed. I was just testing that. Thanks for letting me know.

config/locales/api.en.yml
kammerer: I've already fixed it.

app/views/submissions/steps/trial/_step02.html.haml
Nuanda: I'm sorry for stating the obvious. I'll try to have a look at it, unless you do that already?

kammerer: No, I have not investigated any further.

2016-05-29 12:54 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> In app/views/submissions/steps/trial/_step02.html.haml
> <https://github.com/eSpectrum-IT/brassica/pull/516#discussion_r65001831>:
>
> >
> >  .form-group
> >    = content.label :trait_descriptor_list
> > -  - options = options_for_submission_select(@content.existing_trait_descriptors, :descriptor_name)
> > -  - options.concat options_for_submission_select(@content.new_trait_descriptors, :descriptor_name, id: false)
> > -  = content.select :trait_descriptor_list, options.html_safe, {},
> > +  -#- options = options_for_submission_select(@content.existing_trait_descriptors, :descriptor_name)
> > +  -#- options.concat options_for_submission_select(@content.new_trait_descriptors, :descriptor_name, id: false)
>
> I'm sorry for stating the obvious. I'll try to have a look at it, unless
> you do that already?
>
> —
> You are receiving this because you were assigned.
> <https://github.com/eSpectrum-IT/brassica/pull/516/files/0f131c697bf2c02bde1300456dfc12db84cbd4d6#r65001831>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAA2ye4txVRR9OBDtZi_zsTTWpB8f7Raks5qGXBogaJpZM4IpK03>
> .
>



-- 
Tomasz Szymczyszyn

app/views/submissions/steps/trial/_step02.html.haml
```diff
@@ -9,9 +9,9 @@
 
 .form-group
   = content.label :trait_descriptor_list
-  -#- options = options_for_submission_select(@content.existing_trait_descriptors, :descriptor_name)
-  -#- options.concat options_for_submission_select(@content.new_trait_descriptors, :descriptor_name, id: false)
-  = content.select :trait_descriptor_list, {}, {},
+  - options = options_for_submission_select(@content.existing_trait_descriptors, :trait_name)
+  - options.concat options_for_submission_select(@content.new_trait_descriptors, :trait, id: false)
+  = content.select :trait_descriptor_list, options.html_safe, {},
```
Nuanda: @kammerer Can you test this version? I added both an old and a new TD, moved one step further, moved one step back (both were displayed), removed the old TD (the grey one), moved one step further, moved one step back, the new TD was still displayed.

kammerer: It works now. In my test I forgot to add `.html_safe`...

spec/models/trait_descriptor_spec.rb
```diff
@@ -2,129 +2,39 @@
...
+      create_list(:trait_score, 2, trait_descriptor: td2)
       table_data = TraitDescriptor.table_data
-      expect(table_data.count).to eq 3
-      expect(table_data.map{ |td| [td[2], td[3], td[5]] }).to match_array [
-        [td1.descriptor_name, pts[0].project_descriptor, '1'],
-        [td1.descriptor_name, pts[1].project_descriptor, '2'],
-        [td2.descriptor_name, pts[0].project_descriptor, '2']
+      expect(table_data.count).to eq 2
+      expect(table_data.map{ |td| [td[1], td[6]] }).to match_array [
+        [td1.trait.name, 3],
```
kammerer: Minor thingie - since `TraitDescriptor#trait_name` is defined, I think it should be preferred here.

kammerer:
It might be worth it to consider using [ignorable](https://github.com/nthj/ignorable) gem for table columns left for various reasons but not actually expected to be used anymore.

kammerer:
@Nuanda The downloadable scoring template does not show the name of a new descriptor added in 2nd step (even though it shows correct numer of columns).

app/services/submission/plant_trial_finalizer.rb
```diff
@@ -26,6 +26,12 @@ def create_new_trait_descriptors
     @new_trait_descriptors = (submission.content.step02.new_trait_descriptors || []).map do |attrs|
       attrs = attrs.with_indifferent_access
       attrs = attrs.merge(common_data)
+      if attrs[:trait].present?
+        attrs[:trait] = Trait.find_by_name(attrs['trait'])
+      end
+      if attrs[:plant_part].present?
```
kammerer: Not selecting a PP results in `attrs[:plant_part]` being equal to `""` which causes error on [creation of TD](https://github.com/eSpectrum-IT/brassica/pull/516/files#diff-40e0edd17f0d0745f6ef8476473e2c32R35) (just below).

Nuanda: BTW is there a mechanism to do the above "by ids"? I think you were doing it like that somewhere else in the submissions. That is, to have the submission forms display `trait.name` and `plant_part.plant_part`, where needed, but to internally use `trait.id` and `plant_part.id` - so the finders above do not rely on certain column uniqueness.

kammerer: It should be possible. I will try to change it in such way.

kammerer:
Fixed template generation issue and finalization issue when using Ts of the form `<number>-<string>`.

kammerer:
I changed PPart select to work with ids. Doing the same with Trait is more complicated than I thought though, so I suggest to merge as is and maybe create an issue reminding to improve it later.

Nuanda:
@kammerer I understand the PP issue went away with your last plant_part -> id change. Have you fixed the CSV template not showing the new trait descriptor trait_name in the header? Seems to work correctly now.

kammerer:
Yes, I fixed the template (issue with `str.to_i == 0`). It seems to me that everything works correctly right now.

Nuanda:
@kammerer Thanks for that review!

Nuanda:
Now it also fixes #415 and fixes #423

app/assets/javascripts/data_grid_specific_columns.js.coffee
```diff
@@ -215,6 +215,20 @@ window.configs =
               '&query[plant_scoring_units.plant_trial_id]=' + full[8] + '">' + data + '</a>'
           else
             ''
+      ,
+        targets: 'traits_name_column'
+        render: (data, type, full, meta) ->
+          if data && full[7] && full[7].indexOf("TO:") > -1
+            '<a href="http://browser.planteome.org/amigo/term/' + full[7] + '" target="blank">' + data + '</a>'
```
kammerer: It's `_blank` not `blank`.

Nuanda: Right. Fixed.

nowakowski:
@Nuanda Shouldn't `Trait` be Publishable given that `TraitDescriptor` is?

db/migrate/20160525214823_promote_plant_parts_to_trait_descriptors.rb
```diff
@@ -0,0 +1,10 @@
+class PromotePlantPartsToTraitDescriptors < ActiveRecord::Migration
```
nowakowski: @Nuanda It seems to me that this migration and the one preceding it are two halves of the same apple, i.e. they jointly accomplish an operation which is meant to be atomic. Accordingly, I would suggest merging this migration with `20160525172425_create_traits.rb`.

db/migrate/20160525214823_promote_plant_parts_to_trait_descriptors.rb
```diff
@@ -0,0 +1,10 @@
...
+  def change
+    add_reference :trait_descriptors, :plant_part, index: true
+    add_foreign_key :trait_descriptors, :plant_parts, on_delete: :nullify, on_update: :cascade
+
+    Rake::Task['curate:promote_plant_parts_to_traits'].invoke
+
+    remove_reference :plant_scoring_units, :plant_part
```
nowakowski: While we are indeed dropping a column here, this doesn't have to be irreversible - the information is preserved in the DB and could be recovered. A `down` method could recreate the reference column and then find the required `plant_part` for each PSU by following `trait_scores.first.trait_descriptor.plant_part` before dropping the "new" relationship (i.e. `trait_descriptor` > `plant_part`).

db/migrate/20160525161622_add_label_to_plant_parts.rb
```diff
@@ -0,0 +1,19 @@
...
+    PlantPart.reset_column_information
+    Rake::Task['obo:plant_parts'].invoke
+  end
+
+  def down
+    remove_column :plant_parts, :label
+    remove_column :plant_parts, :canonical
+
+    PlantPart.all.each do |pp|
+      pp.destroy if pp.plant_scoring_units.count == 0
```
nowakowski: This will not work because your branch removes `has_many :plant_scoring_units` from `PlantPart` (you will trigger something like `undefined method `plant_scoring_units' for #<PlantPart:0x00000008916e20>`). Tip: use `column_exists?` to check for the presence of the reference column before trusting ActiveRecord attributes.

Nuanda:
`Trait` should not be Publishable, as it's public by default. It's a vocabulary, hence not "owned" by anyone particular.

db/migrate/20160525214823_promote_plant_parts_to_trait_descriptors.rb
Nuanda: I think you might have misread it - the former one is about _Trait_ Ontology, this one is about _Plant_ Ontology. Two independent things. I did them in a single PR simply because it was convenient implementation-wise.

db/migrate/20160525214823_promote_plant_parts_to_trait_descriptors.rb
Nuanda: You are correct. Done.

db/migrate/20160525161622_add_label_to_plant_parts.rb
Nuanda: Fixed.

db/migrate/20160525214823_promote_plant_parts_to_trait_descriptors.rb
nowakowski: Okay; it's not a serious concern either way.

PR #522, opened 2016-06-02
Accessions in trial submission

### LINKED ISSUE #481: Extend Trial submission to consider Plant Accessions
Wiktor's initial idea was to use RIPR data to test _API submission_ (not the manual submission process) - hence we plan to develop a script to do that, see #394. During a teleconference discussion, between me, @wjurkowski and @nowakowski, we decided that a certain extension to CS PlantAccession model is needed - see #392 (I will ask Piotr to comment on that in more detail below).

@teatree1212 Annemarie suggests, and I think she is right, that a similar possibility should be inbuilt in the _manual_ Trial submission - i.e. to be able to influence how PlantAccessions are being created in the process. Let's use this issue discussion thread to decide what needs to be done to enable that.

Now, the current BIP data model assumes that:
 - each TraitScore is related to exactly one PlantScoringUnit (PSUs can have multiple TSes)
 - each PSU may be related to one PlantAccession (PAs can have multiple PSUs)
 - each PA may be related to a PlantLine.

As you know, PLs are submitted in the other submission process (Population submission). I asked whether PAs should be submitted there as well, or should they be submitted in the Trial submission process - this stays undecided, I think, but I understand we lean towards the second solution.

Also, currently, the user only supplies the PSU name (the first column of the file uploaded in the 3rd step of the Trial submission process) - we should probably support more PSU columns. For convenience I paste the current PSU database model below (without relations):

```
    t.text     "scoring_unit_name" # This is already set by the user
    t.text     "number_units_scored"
    t.text     "scoring_unit_sample_size"
    t.text     "scoring_unit_frame_size"
    t.date     "date_planted"
    t.text     "described_by_whom"
    t.text     "comments"
    t.text     "entered_by_whom" # This is already set, to the user's name, by BIP
    t.date     "date_entered" # This is already set, to the current date, by BIP
    t.text     "data_provenance"
    t.text     "data_owned_by"
    t.text     "confirmed_by_whom"
```

So, submission of values for (some of) these columns might be supported as further columns in the file file uploaded in the 3rd step of the submission. But it's up to you if these columns make sense.

Going back to PAs. From what I understood during the mentioned teleconference about RIPR data, a single Plant Trial may involve a lot of PAs, so it is probably not feasible to ask the user to manually create those PA records by the means of a web form (like it is done for e.g. new Trait Descriptors, in the 2nd step). If this is correct (?), we should probably extend the 3rd step's file definition even further, by introducing more columns which would describe a PA that a particular PSU (remember - we have 1 PSU per file row) is related to.

Then, we have at least 2 issues to solve:
 - how users identify PAs that are already existing in BIP? (probably -> by name)
 - what even further columns a user should supply in the file, in case one wants a new PA record to be created?

@nowakowski Piotr, could you explain what new columns, related to accession identifiers, we were planning to add to the PA table, and give a sample of values from the RIPR data? I think this is valuable for this discussion.

### LINKED ISSUE #519: Spelling mistake Plant scoring trial submission
screen shot 2016-06-01 at 14 45 26

Change "Metarials" to "Materials" in Step 2 Plant scoring trial submission

### LINKED ISSUE #520: Link PSU name column in Trait Scoring pivot table to PSU record in PSU table

Nuanda:
Adds 2 PlantAccession-related fields to the scoring CSV and manages uploaded information accordingly.

Fixes #481 

Also fixes two other minor issues related to PlantTrials. Fixes #519 fixes #520


Nuanda:
This is superseded by #535

Issue #499, opened 2016-05-10
Management of subsequent years of plant trials


Nuanda:
See #488 for the beginning of this discussion (points 1. and 8.).

@teatree1212 So, my proposition, if not decided otherwise by the experimentalists, is to keep all plant trial submission annual. Therefore we will:
 - keep asking for only a single year in the 1st step of the submission
 - will ask for a single image of a plot overview, reflecting the way it was planted this exact year of the trial
 - obviously, scoring should be from this single year only
 - multi-year trials well end up with several records in the PlantTrial table, one per year.

So, the crucial question, of course, is: are there trait scoring scenarios, esp. for mean data, that average over the time axis? Like - there is a single scoring sheet that has e.g. mean values over three consecutive years? Because submission of such data might be problematic having the above assumptions.

teatree1212:
sent out this question to experimentalists.

teatree1212:
One experimentalists opinion sent via email. this tends to support your suggestion.

teatree1212:
Other experimentalists opinion sent via email. i find it supports your suggestions, too. How about, we assume that such mean data can not exist for now and make people submit trials from individual years/+sites. I think even if it exists, it would be better to make the user submit the "rawer" data, including the consecutive years. This is also what the second experimentalist sais; that she would prefer more information about the individual sets of conditions.

**Hence, please go ahead and do as you suggest**. And ignore the unlikely fact that year could be used to construct an average.
In the documentation, I need to include the fact that there is no option to submit this and recommend to submit the yearly trials one by one.

Nuanda:
@teatree1212 One more thing: do you think we should ask the user to submit that image with experiment layout as the final step (i.e. just before the "really final" annotations/comments step)? That is, after the scoring upload? Or after the PlantTrial was defined (directly after step 1)? I guess it is an arbitrary thing, to some extent, just asking your opinion.

wjurkowski:
Perhaps the advantage of the image attachments at the very last step is increasing the chance to have it linked with correct submission to BIP and no time is waisted?

teatree1212:
thinking in where it conceptually makes the most sense for the experimentalist to submit it is when they think about their experimental setups ( in the new step 2.5). 
But 2.5 is already getting quite a complex exercise for the users, so maybe having it before the "really final"  step (as optional), may be the best way to go.

Nuanda:
So, the new Trial submission process would be as follows:
 1. new trial data
[...]

PR #524, opened 2016-06-02
Trial layout image submission

kammerer:
Related to #499. Resolves most issues but does not add a new step to the submission process yet. Layout upload is added to the last step. It will be moved to the proper step in a separate PR.


Nuanda:
Makes sense. I've just been thinking about adding both new steps in a single PR, so this terrible part of work is done in a single, swift, brutal move :).

Is this RtR?

kammerer:
Yes, I forget about adding label.

Nuanda:
What type of files are accepted?

Nuanda:
@kammerer Tomasz, while I review the PR please have a look at my latest commit here - this is a try at showing the picture link in the PT data table. A bit complex but I didn't know how to construct the paperclip attachment directly to `public`, from javascript, having only the DB column data. Is there a better way not involving "decorating" each JSON element of the datatable response with a call to `layout.url`?

kammerer:
I think it is a good approach to show the pictures. Another option would be to store computed urls in the database and include them in JSON.

Nuanda:
I have added some more text. Also, a thumbnail href fix.

app/assets/javascripts/submissions.coffee
```diff
@@ -270,7 +270,7 @@ class TrialSubmission extends Submission
         @$('.uploaded-layout .delete-layout-upload').attr(href: data.result.delete_url)
 
         @$('.uploaded-layout .layout-image').prop(src: data.result.small_file_url)
-        @$('.uploaded-layout .layout-image').parent().prop(href: data.result.original_file_url)
+        @$('.uploaded-layout .layout-image').parent('a').prop(href: data.result.original_file_url)
```
kammerer: This change looks suspicious - was it needed? AFAIR `parent()` traverses one level up, if you pass selector as an argument it returns empty set unless parent matches that selector.

Nuanda: Well, it didn't work without it. That's why I did that fix.

kammerer: It definitelly works for me without this change.

2016-06-04 16:06 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> In app/assets/javascripts/submissions.coffee
> <https://github.com/eSpectrum-IT/brassica/pull/524#discussion_r65802931>:
>
> > @@ -270,7 +270,7 @@ class TrialSubmission extends Submission
> >          @$('.uploaded-layout .delete-layout-upload').attr(href: data.result.delete_url)
> >
> >          @$('.uploaded-layout .layout-image').prop(src: data.result.small_file_url)
> > -        @$('.uploaded-layout .layout-image').parent().prop(href: data.result.original_file_url)
> > +        @$('.uploaded-layout .layout-image').parent('a').prop(href: data.result.original_file_url)
>
> Well, it didn't work without it. That's why I did that fix.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/eSpectrum-IT/brassica/pull/524/files/8577b3a75c449dcc2041b914f48f772ef38128da..4af566e069cb847caf6622cff3018785129d7bb7#r65802931>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAA2yeO3zZh16vOC0v_yq7_h0FoSYV2xks5qIYZ3gaJpZM4Is_Pt>
> .
>



-- 
Tomasz Szymczyszyn

Nuanda:
Another request - please add the link to the layout file, if present, in submission show.

app/assets/javascripts/submissions.coffee
Nuanda: All right, reverted that. I definitely had the problem of the href not being updated when I added the image, yet it looks like working ok now. I guess it was something else.

Nuanda:
The third request - could you add some specs for the finalizer saving the uploaded picture as a PT attribute?

kammerer:
Updated PR

app/decorators/plant_trial_submission_decorator.rb
```diff
@@ -129,4 +129,16 @@ def data_provenance
   def comments
     object.content.step04.comments.presence
   end
+
+  def layout_url
+    if object.submitted_object.try(:layout).present?
+      h.plant_trial_path(object)
```
Nuanda: Shouldn't that be `object.submitted_object`?

Nuanda:
Apart from the `object.submitted_object` issue, it's :+1: and ready to merge.

app/decorators/plant_trial_submission_decorator.rb
kammerer: Yes, of course.

PR #529, opened 2016-06-03
Extend PlantAccession model with an optional link to PlantVariety

### LINKED ISSUE #500: Add PlantVariety relation to PlantAccession
As decided in #481 - please add these two relations and make them mutually exclusive, i.e., no single PSU could be related to both a PL and a PV at the same time.

I'm not sure what to do with the current PSUs. I guess we can safely assign PL relation to those PSUs that have a PA relation and the related PA has a PL assigned.

How many PSUs would be left with no PL after that?

nowakowski:
Fixes #500


Nuanda:
:+1:

PR #534, opened 2016-06-07
Fix bug which appears when querying certain models by publishable params

### LINKED ISSUE #533: Solve the double-join issue in table_data when filtering
Found for QTLs, but probably present for other models as well.

nowakowski:
Fixes #533


Nuanda:
:+1:

PR #535, opened 2016-06-07
Extension to plant trial submission step3

### LINKED ISSUE #407: Providing content for better explanation messages (context help) for submissions
It was suggested that the current submission steps should be outfitted with better help messages, more explanations, examples etc. Please provide these text messages so we may put them in the right place inside the submission processes.

### LINKED ISSUE #481: Extend Trial submission to consider Plant Accessions

### LINKED ISSUE #519: Spelling mistake Plant scoring trial submission

### LINKED ISSUE #520: Link PSU name column in Trait Scoring pivot table to PSU record in PSU table

### LINKED ISSUE #523: Support reading and recording technical replicate numbers in trial submission
Related: #525 #526

### LINKED ISSUE #525: Extend Trial Scoring pivot table to show technical replicates

### LINKED ISSUE #526: Show technical replicate number in Trait Scores data table

Nuanda:
Included:
 - plant accessions (parsing in parser, finding/creating in finalizer, publishing in publisher, showing in generated template)
 - technical replicates (parsing in parser, creating as TraitScores in finalizer, showing in generated template)
 - design factors (parsing in parser, creating in finalizer, showing in generated template)
 - plant varieties/plant lines (parsing in parser, creating/assigning in finalizer, publishing in publisher, showing in generated template)

Issues:
 - fixes #481 
 - fixes #523 
 - fixes #525 
 - fixes #526
 - fixes #407
 - also fixes #519 (a minor issue)
 - also fixes #520 (a minor issue)


Nuanda:
@kammerer I merged master and I'm in the process of fixing specs. Can we proceed with #538 directly in here?

app/controllers/submissions/uploads_controller.rb
```diff
@@ -6,12 +6,33 @@ class Submissions::UploadsController < ApplicationController
   def new
     @traits = PlantTrialSubmissionDecorator.decorate(submission).sorted_trait_names
 
+    # TODO FIXME Add technical replicates generation
+    pl_pv_name = if submission.content.step03.lines_or_varieties == 'plant_varieties'
```
kammerer: I'd suggest moving template generation to a separate class.

Nuanda: You mean a Service class?

app/controllers/submissions/uploads_controller.rb
```diff
@@ -4,14 +4,51 @@ class Submissions::UploadsController < ApplicationController
...
+                   'line'
+                 end
+
+    design_factor_names = submission.content.step03.design_factor_names || []
+    design_factors = {
+      'A' => design_factor_names.map{ 1 },
+      'B' => design_factor_names.map{ 1 }
+    }
+    if design_factors['B'].present?
+      design_factors['B'] = design_factors['B'][0..-2] + [2]
```
kammerer: Simpler code: `design_factors['B'][-1] = 2`

Nuanda: Fixed.

app/controllers/trial_scorings_controller.rb
```diff
@@ -27,7 +29,10 @@ def show
   private
 
   def prepare_grid_data
-    objects = @plant_trial.scoring_table_data(trait_descriptor_ids, current_user.try(:id))
+    @replicate_numbers = replicate_numbers
```
kammerer: I believe this assignment is not needed.

app/controllers/submissions/uploads_controller.rb
kammerer: Yes.

app/decorators/submission_trait_scores_upload_decorator.rb
```diff
@@ -12,14 +12,47 @@ def parser_summary
...
 
-        summary << " - parsed #{trait_scores.size} plant(s) with unique identification"
+        summary << " - parsed #{trait_scores.size} plant scoring unit(s) with unique identification"
         histogram.each do |scoring_number, plants|
-          summary << "  - #{plants.size} plant(s) have #{scoring_number} trait score(s) recorded"
+          summary << "  - #{plants.size} unit(s) have #{scoring_number} trait score(s) recorded"
         end
       end
 
+      accessions = object.submission.content.step04.accessions
```
kammerer: It's not easy to reason about the result of this method. Have you thought about using erb template to separate data preparation and formatting?

app/forms/submissions/trial/step03_content_form.rb
```diff
@@ -1,8 +1,9 @@
 module Submissions
   module Trial
     class Step03ContentForm < PlantTrialForm
-      # NOTE: needed because completely empty step breaks submission
-      property :_i_am_fake
+      property :lines_or_varieties, writeable: false
```
kammerer: What does `writeable: false` do?

app/forms/submissions/trial/step04_content_form.rb
```diff
@@ -1,15 +1,25 @@
...
       property :trait_scores, writeable: false
+      property :accessions, writeable: false
+      property :lines_or_varieties, writeable: false
+      property :replicate_numbers, writeable: false
+      property :design_factor_names, writeable: false
+      property :design_factors, writeable: false
       property :upload_id
-      property :trait_mapping, writeable: false
 
       def self.permitted_properties
```
kammerer: It does not seem to be of any use to override `permitted_properties` here as it returns all defined properties anyway.

app/services/submission/plant_trial_finalizer.rb
```diff
@@ -35,10 +35,80 @@ def create_new_trait_descriptors
...
     trait_mapping = submission.content.step04.trait_mapping
+    trait_scores = submission.content.step04.trait_scores || {}
+    accessions = submission.content.step04.accessions
+    replicate_numbers = submission.content.step04.replicate_numbers || {}
+    design_factors = submission.content.step04.design_factors || {}
+    design_factor_names = submission.content.step04.design_factor_names || []
+    lines_or_varieties = submission.content.step04.lines_or_varieties || {}
+
+    @new_plant_scoring_units = trait_scores.map do |plant_id, scores|
+      rollback(2) unless accessions[plant_id] && accessions[plant_id]['plant_accession'] && accessions[plant_id]['originating_organisation']
```
kammerer: How is that possible? Doesn't parser ignore lines with missing accession? In case it is due to a user error - will user be informed explicitly what was wrong? If that's a guard for really unexpected situation then we should probably just fail.

app/services/submission/plant_trial_finalizer.rb
```diff
@@ -35,10 +35,80 @@ def create_new_trait_descriptors
...
+          if plant_line
+            accession_relation = { plant_line: plant_line }
+          else
+            #TODO Inform the user there is no PlantLine of a given name and that it should be submitted with another submission first
+            rollback(2)
+          end
+
+        else
+          # Unknown and not expected class name used for 'relation_class_name'
+          rollback(2)
```
kammerer: Again, it does not seem possible.

app/services/submission/plant_trial_finalizer.rb
```diff
@@ -35,10 +35,80 @@ def create_new_trait_descriptors
...
+        factor_attrs = {}
+        factor_values.each_with_index do |factor_value, i|
+          factor_attrs["design_factor_#{i+1}".to_sym] = if design_factor_names[i]
+                                                          "#{design_factor_names[i]}_#{factor_value}"
+                                                        else
+                                                          nil
+                                                        end
+        end
+        design_factor = DesignFactor.create!(
+          common_data.delete_if{ |k,v|
```
kammerer: `common_data.except(:user, :published, :published_on)`

app/services/submission/plant_trial_finalizer.rb
```diff
@@ -47,10 +117,19 @@ def create_scoring
           trait_descriptor = get_nth_trait_descriptor(trait_mapping[col_index])
           rollback(1) unless trait_descriptor
 
+          trait_score_attributes = common_data
```
kammerer: Not sure if it is a problem here but `trait_score_attributes = common_data.dup` would be safer. Otherwise you may unwillingly assign `technical_replicate_number`.

app/services/submission/plant_trial_finalizer.rb
```diff
@@ -35,10 +35,80 @@ def create_new_trait_descriptors
 
   def create_scoring
```
kammerer: ~100 lines for a single method seems a bit too much ;)

app/services/submission/trait_score_parser.rb
```diff
@@ -24,32 +30,72 @@ def call
...
+      @upload.log "Interpreting design factors" if @number_of_design_factors > 0
+      header[1, @number_of_design_factors].each do |design_factor_name|
+        @design_factor_names << (design_factor_name ? design_factor_name.gsub('/','') : '')
+      end
+
+      @line_or_variety = 'PlantVariety' if header.index('Plant line').nil?
+
+      @upload.log "Mapping file header columns to Trait Descriptors"
+      replicates_present = detect_replication(header[4..-1])
+      header[4..-1].each_with_index do |column_name, i|
```
kammerer: Shouldn't that start with `4 + @number_of_design_factors`?

app/forms/submissions/trial/step03_content_form.rb
Nuanda: Does not persist this value in associated active record in the DB. However, you we probably don't need it here as you do not "associate" that form with any particular AR class, right? I mean, you use "reform" in a slightly different way, correct?

app/forms/submissions/trial/step04_content_form.rb
Nuanda: I kind of automatically assumed that no override == no properties are permitted. Is that not the case?

app/services/submission/plant_trial_finalizer.rb
Nuanda: Yes, this "should never happen(tm)". I simply didn't want the situation when our broken code allows submission of some incorrect data. Should I rather simply throw an uncaught exception (500)?

app/forms/submissions/trial/step03_content_form.rb
kammerer: That's right. I think in our case it does nothing as all properties are passed to `Submission#content` (otherwise they would be useless, wouldn't they?)

app/forms/submissions/trial/step04_content_form.rb
kammerer: No, all properties are permitted by default: https://github.com/eSpectrum-IT/brassica/blob/master/app/forms/base_form.rb#L5

app/services/submission/trait_score_parser.rb
Nuanda: Probably didn't matter but you are right.

app/forms/submissions/trial/step03_content_form.rb
Nuanda: I admit I'm not that familiar with reform, so I simply tried to be (too) safe here. But does that mean that we can also remove that constraint from e.g. data_status in step 1?

app/services/submission/plant_trial_finalizer.rb
kammerer: I think that will be better because:

- we will be notified that sth went wrong by email
- user will not be left with submission failing for some hard to guess reason (probably a bug)

Optionally, if we want to be extra nice as this is an important step, we could provide custom error message for the action of finalizing submission, so that we ensure user that the data is safe and we will look into this issue.

app/forms/submissions/trial/step03_content_form.rb
kammerer: Yes, it can be removed from all the properties.

app/services/submission/trait_score_parser.rb
```diff
@@ -82,4 +149,22 @@ def csv
...
   end
+
+  def detect_replication(header_columns)
+    header_columns.any?{ |column_name| column_name && column_name.index(/rep\d+$/) }
+  end
+
+  def split_to_trait_and_replicate(column_name)
+    replicate_number = 0
+    return ['',0] if column_name.blank?
+    trait_name = column_name.strip.gsub(/rep\d+$/) do |replicate|
```
kammerer: It took me a while to see what it does :) How about that?
```ruby
matches = column_name.strip.match(/\A(?<trait_name>.+)_rep(?<replicate_number>\d+)\z/)
[matches[:trait_name], matches[:replicate_number]]
```

BTW Doesn't the original code leave trait names suffixed with an underscore?

app/services/submission/plant_trial_finalizer.rb
```diff
@@ -35,10 +35,80 @@ def create_new_trait_descriptors
...
+      plant_accession = PlantAccession.find_or_create_by!(
+        plant_accession: accessions[plant_id]['plant_accession'],
+        originating_organisation: accessions[plant_id]['originating_organisation']
+      ) do |new_plant_accession|
+        # If new accession, set PL or PV
+        accession_relation = {}
+
+        if lines_or_varieties[plant_id].blank?
+          # Info on either PL or PV for a new accession is required
+          rollback(2)
```
Nuanda: I actually found an error is specs thanks to that remark ;).

app/views/submissions/steps/trial/_step03.html.haml
```diff
@@ -1,4 +1,50 @@
...
+  %p
+    In case you would like to check what plant accessions are currently registered in BIP, please
+    use the
+    = link_to 'plant accessions table view', data_tables_path(model: 'plant_accessions'), target: '_blank'
+    (you may filter data in that table with the
+    %b Search
+    text field).
+
+  %p
+    If you submit raw data, and you will have an option to submit technical replicate readings as
```
kammerer: `and you will` -> `you will`

Nuanda:
Commited fixes for all remarks with :+1: . I'll try to address the rest of them tomorrow. Thanks a lot for that work.

app/views/submissions/steps/trial/_step04_before.html.haml
```diff
@@ -1,54 +1,118 @@
...
-      assay_09,56.98,249.00
-      assay_08,57.24,252.00
-      assay_98,18.48,255.00
-      assay_97,20.40
-      assay_96,50.82,254.00
-      assay_93,58.50
-      assay_94,,252.00
+    For your convenience please use
+    %b
+      = link_to 'the ready template', new_submission_upload_path(submission)
```
kammerer: This link is not that easy to find in the block of text. Maybe we should add a button below the blue text block, before "Choose a file" button?

app/views/trial_scorings/show.html.haml
```diff
@@ -17,6 +17,10 @@
       %tr
         %th.plant_scoring_units_scoring_unit_name_column
           = t 'tables.plant_scoring_units.scoring_unit_name'
-        - @trait_descriptors.map{ |td| td.try(:trait).try(:name) }.each do |column|
-          %th= column
+        - @trait_descriptors.map{ |td| [td.id, td.try(:trait).try(:name)] }.each do |id, column|
```
kammerer: I know this was not changed now, but it is possible for TD to not have a T, or for a T to not have a name?

app/services/submission/trait_score_parser.rb
Nuanda: I actually like my use of gsub with block better - it is actually simpler in my opinion compared to your version.

However, I did change placeholder values generator - before I opted for no-space, all-underscore placeholders so it is easier to click-select it in a tool like Excel and replace. But I changed my mind in the middle of the process, as you see - now I removed underscores completely. I think the template is easier to comprehend this way.

app/views/trial_scorings/show.html.haml
Nuanda: I think not. Both have `null: false` in the schema and there is the FK constraint on the `TD.trait_id` column, so that should not happen. In worst case scenario, T.name could be an empty string.

app/views/submissions/steps/trial/_step04_before.html.haml
Nuanda: I know and it was reported. So I moved it up in the manual, I made the font bold and I mention it again in the last part of the manual. I really can't imagine anyone properly submitting that data without reading the whole page very carefully first - so they should see this link anyway.

Do you think users will simply not recognize it as something clickable/downloadable?

app/services/submission/plant_trial_finalizer.rb
Nuanda: Well, normally I would never be able to write so long a method in a single stroke. This one grew from addition to addition. Applied naive refactoring ;).

spec/services/submission/trait_score_parser_spec.rb
```diff
@@ -5,101 +5,234 @@
...
-        input_is "id,trait 2  ,\"trait,1\""
-        subject.send(:map_headers_to_traits)
+        input_is "id,Plant accession,oo,Plant line,trait 2  ,\"trait,1\""
+        subject.send(:parse_header)
         expect(subject.trait_mapping).
           to eq({0 => 1, 1 => 0})
       end
 
       it 'honors proper trait sorting by index' do
         upload.submission.content.update(:step02, trait_descriptor_list: ['Ctrait', 'Atrait', 'Btrait'])
```
kammerer: What is the mapping "direction"? The spec would be more expressive, with eg. `trait_descriptor_list: ['Ctrait', 'Atrait', 'Btrait']` and `input_is "id,Plant accession,oo,Plant line,Btrait,Ctrait,Atrait"`

spec/services/submission/trait_score_parser_spec.rb
```diff
@@ -5,101 +5,234 @@
...
         upload.submission.content.update(:step02, trait_descriptor_list: [td.id, 'new trait'])
-        input_is "id,new trait,old trait"
-        subject.send(:map_headers_to_traits)
+        input_is "id,Plant accession,oo,Plant line,new trait,old trait"
+        subject.send(:parse_header)
         expect(subject.trait_mapping).
           to eq({0 => 1, 1 => 0})
       end
 
       it 'reports error on repetitive mapping' do
```
kammerer: I don't get this one - I would expect repeated traits in header.

app/decorators/submission_trait_scores_upload_decorator.rb
Nuanda: Maybe a new partial rendered inside _step04.haml, that uses methods provided by the decorator?

kammerer: That would be good too.

spec/services/submission/trait_score_parser_spec.rb
```diff
@@ -108,13 +241,87 @@
...
+
+    it 'parses accession information to a dedicated structure' do
+      input_is "plant 1,pa,oo,pl\nplant X,pa\nplant Y\nplant 2,pa2,oo,pl"
+      subject.send(:parse_scores)
+      expect(subject.accessions).
+        to eq({ 'plant 1' => { plant_accession: 'pa', originating_organisation: 'oo' },
+                'plant 2' => { plant_accession: 'pa2', originating_organisation: 'oo' }})
+    end
+
+    it 'ignores scores beyond number of trait columns' do
```
kammerer: Shouldn't that also give a warning? Clearly, something is wrong with the uploaded data.

app/decorators/submission_trait_scores_upload_decorator.rb
Nuanda: Would I be able to render it to string in the as_json method of the decorator? I think rendering to string is possible only in controllers, right?

app/services/submission/trait_score_parser.rb
kammerer: Come on, you needed a comment to explain what it does and my version automagically teaches the reader about named capture in regexp ;) Seriously though, it's a minor thingie, I don't mind leaving it as it is.

app/decorators/submission_trait_scores_upload_decorator.rb
Nuanda: I decided to refactor that in-situ. The partial would have half of its code asking the decorator for data anyway.

Nuanda: In future refactoring we should come up with a decent solution for things like `object.submission.content.step04.trait_scores` - maybe using PTS decorator everywhere, and putting these step-number-dependent calls there.

spec/services/submission/trait_score_parser_spec.rb
Nuanda: Yes, this maps column ordinal numbers (starting with 0 for the first scoring column). There should be two cases supported:
 - mapping in any order as long as the user keeps the exact trait names
 - mapping in exact trait order if the user does not do that (and you know - the encoding hell... ;)).

Tech reps, however, are expected in natural order anyway (I don't think anyone would start with rep2 in order to follow with rep1 for the same trait...).

spec/services/submission/trait_score_parser_spec.rb
Nuanda: This is trickier. When you read my previous comment, you'll know there are two strategies to that. And in this case the user is somehow mixing them together - s/he puts some completely different named trait in place, where Atrait scores should be given, yet s/he puts scores for Atrait elsewhere. This results in ambiguity. We should ask the user to follow one approach: either column headers with exact trait names, or consistent column ordering.

I try to explain it to the user in the manual, and the template should be correct anyway, unless the user shuffles column headers for some reason.

Nuanda:
All remarks addressed in one way or another. I see two approaches now:
 - you add the forms needed in step3 and we test everything thoroughly again, including manual UI tests, in this PR (I prefer this approach)
 - or I do the manual testing as it is (I'll try to use the console to set form properties for step3) now, we merge it to master, and then you create a separate PR for the forms.

What do you think @kammerer ?

app/decorators/submission_trait_scores_upload_decorator.rb
kammerer: Ok, that refactoring will do.

I did some research regarding rendering partials to string, so I will share anyway. It seems it's not that easy to do that in a decorator unless you pass controller as the [context](https://github.com/drapergem/draper#adding-context). Otherwise, you can use `ActionView::PartialRenderer` directly but I am not sure if it is considerer public API, probably not (and it needs somewhat magical arguments).

```ruby
class Decorator < Draper::Decorator
  def render_errors
    ActionView::PartialRenderer.new(h.lookup_context).render(h, { partial: "/errors", locals: { errors: { :foo => "foo!" } } }, nil)
  end
end
```


Another option would be to use erb or haml directly.

```ruby
engine = Haml::Engine.new("%p Haml code!")
engine.render #=> "<p>Haml code!</p>\n"
```

Nuanda: Yes, I read about the other one as well. Involves directly reading the partial file to a string - a strange solution... In such a case we should probably rewrite the controller action to render the partial itself, passing the decorated upload in locales.

kammerer: I tried to "flatten"  `Submission#content` structure as an experiment, so that instead of `content.step04.trait_scores` we would write `content.trait_scores`, but I am not sure it is worth it. What do you think? The downside (or maybe not?) is that we would need to be more explicit about data we use to create some objects in finalizer. Eg. right now we use all data from the first step for PT `attrs = submission.content.step01.to_h.merge(common_data)`.

Nuanda: What I (really ;)) meant was "Let's not do that right now, as we have a lot of other stuff to write/test, but in case we are asked for considerable submission process extensions in the future, we should address it somehow.".

If Wiktor's plans work (we hope they do) and BIP starts to be widely used by plant experimentalists, we will most probably go back to that issue some day.

kammerer: Yeah, my comment was just FYI. I know that we will not be able to refactor everything right now (or ever ;).

kammerer:
I am OK with adding forms to this branch.

I am done with the review. There is a lot of changes here, so I might have missed something. I am somewhat hesitant about that, as I am not the one putting most effort here, but I felt like I should comment about the structure of the code as the app is growing and I believe it is worth to keep the source as readable as possible.

kammerer:
So, :+1: for the present changes.

Nuanda:
All right. I think we might still do some work to the present parts after you add the forms and we click-test everything together again.

Issue #502, opened 2016-05-11
Design of interactive scoring data template generation


Nuanda:
@kammerer Tomasz, as discussed in #488, there is an idea to introduce what we call 'step 2.5' of the Trial submission. This step would be a form which asks the user a set of questions, and then creates a CSV template, to be later used by the user to submit scoring data (current step 3).

Do you think that we should extend the number of steps to 5, in order to accommodate that, or could we simply start the 3rd step with the form, and then create the template and show the upload button? That would make this step having three individual "substages", but all of them closely related.

teatree1212:
I will design a template based on standard experimental designs. it may be similar to this one used for SRA sequence submissions:
[SRA_BioSample_template.xlsx](https://github.com/eSpectrum-IT/brassica/files/263131/SRA_BioSample_template.xlsx)

teatree1212:
see draft here
https://docs.google.com/spreadsheets/d/1hg4yUOGzzYZ5Yolx02dQ4OBBwQERAQU1gmoo10ZgfoQ/edit#gid=0

Nuanda:
@kammerer What @teatree1212 designs here will eventually be the output of the CSV template generation process. But please tell me what you think about the original question (the wizard/UX side to the issue).

kammerer:
I think adding another step would probably be easier than step 2.5 (which would require more custom javascript code). It should not be a big problem to change the number of steps only for trial submission.

Nuanda:
@kammerer Thanks. I tend to agree with that option as well. Let's wait a bit with that task, since we still don't know if accessions are required or not.

kammerer:
@Nuanda If we do the work required to vary number of steps for different submission types then I would also suggest splitting the first step and putting location-related fields on a separate page. The first step is big and there are required fields in two places quite far from each other.

Nuanda:
@kammerer I'd rather not. True, it is a bit long, but all that is logically related.

Nuanda:
@teatree1212 Before we go with the design matrix "2.5 step", please have a look at the following columns in the PlantTrial table:
  - `design_type`, with values like:
    - "randomized blocks"
    - "incomplete block design"
    -  "alpha design"
    - "randomised complete block design with restricted blocking for spring types"
    - "randomised complete block design"
  - `statistical_factors`, with values like:
    - "occasion / bench / block / plot / pot"
    - "occasion / replicate / block / plot / pot"
    - "replicate / area-pair / block / plot"
[...]

PR #536, opened 2016-06-08
Variable number of submission steps

### LINKED ISSUE #499: Management of subsequent years of plant trials

kammerer:
Fixes #499 and related to #502.

- it is now possible to have different number of steps in submissions of different types
- trial layout image upload is extracted into its own step 
- empty placeholder step is added between traits and scoring 

Note that it probably should not be merge to master with this empty step added.


Nuanda:
I've got my own branch that has almost all needed additions to parser/finalizer (not to forms). Of course, this PR impacts that branch. How do you propose to proceed with merging (after this one is reviewed)?

Nuanda:
@kammerer Tomasz, I tried to update step count for things related to Trial Submission. Please see the changes in case I did something wrong.

One thing that certainly is still wrong is the publishable/published/visibility migration, that needs to be a bit smarter now. I added a TODO there in the code - please try to fix that.

kammerer:
Argh, I see now that PR was somewhat incomplete... I started working on "flattening" submission content, so that instead of `submission.content.step06.data_provenance` we would just call `submission.content.data_provenance`, but then I got doubts and skipped that change. As a result I forgot to update step numbers in many places.

kammerer:
As for the migration - what do you think about removing the data-migration part altogether? I assume it is already deployed, so there is no point in keeping that code forever.

There is another issue that I forgot to mention - the contents of existing submissions will no longer be valid, I mean, they will / would crash the finalizer. We need to decide if we want to migrate them or keep without changes. The latter is OK with me unless there are any "real" trial submissions in progress.

Nuanda:
Ok, we will merge this with master, then. We will simply not deploy it to production before we merge the other branch, and then maybe also with new step 3 form.

All right re. the migration. Also, I don't think there are any real submissions in progress right now. We may, however, double check that with our users before deploying (and then we will destroy the unfinished submissions before we change the code base). Are we talking only trial submissions? Or all submissions?

kammerer:
Only trial submissions.

2016-06-08 22:16 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> Ok, we will merge this with master, then. We will simply not deploy it to
> production before we merge the other branch, and then maybe also with new
> step 3 form.
>
> All right re. the migration. Also, I don't think there are any real
> submissions in progress right now. We may, however, double check that with
> our users before deploying (and then we will destroy the unfinished
> submissions before we change the code base). Are we talking only trial
> submissions? Or all submissions?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/eSpectrum-IT/brassica/pull/536#issuecomment-224714121>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAA2yXNiIyojFgboYU0fPPTGOTlIWI0Vks5qJyMogaJpZM4Iwq3Z>
> .
>



-- 
Tomasz Szymczyszyn

Nuanda:
@kammerer So, it's :+1: from me to merge it with master. If you think the changes I did (regarding renumbering of steps here and there) are ok, please go ahead with the merge.

kammerer:
@Nuanda Your changes are fine, thanks. I am not sure about merging to master right away. Shouldn't we wait for your changes and #538 to be ready? If we merge now, deploying sth else in the meantime will be more difficult (still possible of course).

Nuanda:
I don't mind having that temporary in master. #535 is rather finished and awaits that PR. We may merge this branch to #535 if you prefer that. Then, when finished, I will PR the other branch and we will delete this one after successful merge. Would that be better?

We can work together on #538 inside #535, after we merge this branch here.

Issue #450, opened 2016-03-25
Introduce global param sanitizer


kammerer:

teatree1212:
cultivar names definitely need special characters, and other name related fields

kammerer:
@teatree1212 Do you think it is possible to reliably tell which fields should accept special chars and which ones can filter them out (or filter a subset)? If not then perhaps this is not a viable solution.

teatree1212:
special characters in cases of trial descriptions or comments
 a special character is even the Degree (°) character for mentioning a temperature or a set of ' ' to emphasise on a ceratain word -> Can imagine this coming up in 
trial_methods,
 trial description, 
various comments-fields.

other fields with special characters, as they may include names :
place names ( need to use ontology at some point anyways- waiting for ELIXIR to decide on ontology)
plant_variety, (PV)
plant_line_name, (PL)
Plant accessions ( PA [everything is possible..])
contact-person ( all tables)
data_provenance (all tables)
data_owned_by ( all tables)
population_owned_by  (PP)
.._description (PP, PT)
described_by_whom (all tables)
confirmed_by_whom ( all tables)

-> special characters can be anticipated in most of the fields within the database..

PR #537, opened 2016-06-08
Escape HTML in datatable contents server-side

kammerer:
Related to #450.

All json data for datatables is html-escaped server-side (before caching). I've browsed through different datatables after applying the fix and I haven't noticed any issues. It is a partial solution to the injection issue, it does not protect from stored injection though.

### DIFF
--- a/app/controllers/data_tables_controller.rb
+++ b/app/controllers/data_tables_controller.rb
@@ -30,7 +30,16 @@ class DataTablesController < ApplicationController
 
   def prepare_grid_data
     objects = model_klass.table_data(params, current_user.try(:id))
+    html_escape_grid_data(objects)
     ApplicationDecorator.decorate(objects).as_grid_data
+  end
+
+  def html_escape_grid_data(objects)
+    objects.each do |object|
+      object.map! do |value|
+        value.is_a?(String) ? ERB::Util.html_escape(value) : value
+      end
+    end
   end
 
   def model_param


Nuanda:
What kind of processing time overhead have you observed for that all-values loop? Looks a bit scary to me ;) - yes, I know we've got caching... and, one day, maybe server-side filtering, ordering, pagination...

kammerer:
I did some benchmarks on my box (production env, no json caching):

- all plant lines with no escaping
```sh
$ ab -n 100 http://lvh.me:3000/data_tables.json?model=plant_lines
Document Path:          /data_tables.json?model=plant_lines
Document Length:        2711503 bytes

Concurrency Level:      1
Time taken for tests:   102.237 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      271186600 bytes
HTML transferred:       271150300 bytes
Requests per second:    0.98 [#/sec] (mean)
Time per request:       1022.373 [ms] (mean)
Time per request:       1022.373 [ms] (mean, across all concurrent requests)
Transfer rate:          2590.35 [Kbytes/sec] received
```

- all plant lines with escaping
```sh
$ ab -n 100 http://lvh.me:3000/data_tables.json?model=plant_lines
Document Path:          /data_tables.json?model=plant_lines
Document Length:        2711607 bytes

Concurrency Level:      1
Time taken for tests:   138.553 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      271197000 bytes
HTML transferred:       271160700 bytes
Requests per second:    0.72 [#/sec] (mean)
Time per request:       1385.526 [ms] (mean)
Time per request:       1385.526 [ms] (mean, across all concurrent requests)
Transfer rate:          1911.48 [Kbytes/sec] received
```

So, it is considerably slower but I think the impact on users should not be that big due to caching.

Of course, if we decide to scrub user input then it probably does not make sense to do this.

Nuanda:
If we assume all original CS data is injection free, do you think that filtering only objects that have non-nil `user_id` will help a bit?

kammerer:
For sure. But I have another idea, so let's wait with this a bit.

2016-06-23 21:48 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> If we assume all original CS data is injection free, do you think that
> filtering only objects that have non-nil user_id will help a bit?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/eSpectrum-IT/brassica/pull/537#issuecomment-228163147>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAA2yeav66hVD1MDar9bDPQataD5qTd-ks5qOuMNgaJpZM4IxXah>
> .
>



-- 
Tomasz Szymczyszyn

PR #539, opened 2016-06-09
Allow navigate from TD table to PT table with "related" link

### LINKED ISSUE #517: Add link to related Trials in the new TraitDescriptors data table view

Nuanda:
Fixes #517


Nuanda:
I decided to get the IDs array, instead of a simple count, after all - makes the "other side" query much faster (comparing id with an array of IDs, instead of making those joins again, from the other side this time).

nowakowski:
:+1:

PR #542, opened 2016-06-14
Translate outdated seq ids

### LINKED ISSUE #541: Translate outdated GI numbers to Accession.Version for NCBI links
The GI numbers will soon be outdated and not supported anymore:

https://www.ncbi.nlm.nih.gov/news/03-02-2016-phase-out-of-GI-numbers/

This affects links in MLH.bac_hit_seq column. We need to translate them somehow to the NCBI accession numbers.

Nuanda:
Fixes #541


db/migrate/20160614175042_translate_outdated_sequence_identifiers.rb
```diff
@@ -0,0 +1,25 @@
+class TranslateOutdatedSequenceIdentifiers < ActiveRecord::Migration
+  def up
```
nowakowski: It might be a good idea to invoke curate:get_new_sequence_accessions as the first action in `up`.

lib/tasks/translate_ncbi_ids.rake
```diff
@@ -0,0 +1,36 @@
...
+  task get_new_sequence_accessions: :environment do
+    puts "Building mapping for Probe sequences"
+    probes = Probe.where.not(sequence_id: nil).pluck(:id, :sequence_id)
+    probe_accessions = ''
+    probes.each do |id, sequence_id|
+      accession = `curl -s "http://www.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=nucest&rettype=acc&retmode=text&id=#{sequence_id}"`
+      probe_accessions += "#{sequence_id},#{accession}"
+      puts "#{sequence_id},#{accession}"
+    end
+    File.write('db/data/probe_accessions.csv', probe_accessions)
```
nowakowski: Not familiar with this syntax. Are you sure this will overwrite the file, if it exists (which is what we want)? Perhaps it would be safer to always delete both files as the final action in `TranslateOutdatedSequenceIdentifiers.up`?

nowakowski:
Assuming that the rake task is only here for archival purposes and does not actually need to be invoked (since both lists are already bundled with the PR), :+1: .

lib/tasks/translate_ncbi_ids.rake
Nuanda: I think it does (http://ruby-doc.org/core-2.1.0/IO.html#method-c-write).

I don't use it in the migrations anyway.

db/migrate/20160614175042_translate_outdated_sequence_identifiers.rb
Nuanda: I'd rather not since it is lengthy, it is deterministic but, due to network problems, it could cause errors on production migration.

PR #546, opened 2016-06-15
Add widgets for technical replicate numbers and design factor names

kammerer:
- some texts need to be improved (eg. placeholders for design factor names)
- there are no specs yet
- `technical_replicate_numbers` is an array sorted according to `sorted_trait_names`


.rspec
```diff
@@ -1,3 +0,0 @@
---color
```
Nuanda: Give me my console coloring back... :)

app/forms/submissions/trial/step03_content_form.rb
```diff
@@ -2,8 +2,27 @@ module Submissions
...
+          number.to_i.to_s == number && number.to_i > 0
+        end
+        errors.add(:technical_replicate_numbers, :invalid) unless all_valid
+      end
+
+      def technical_replicate_numbers
+        super || []
+      end
+
+      def technical_replicate_nummbers=(numbers)
```
Nuanda: A typo in the name.

app/forms/submissions/trial/step03_content_form.rb
```diff
@@ -2,8 +2,27 @@ module Submissions
...
+        super(numbers.presence || [])
```
Nuanda: In what case do we need `presence` here, instead of just `numbers || []`?

app/decorators/plant_trial_submission_decorator.rb
```diff
@@ -43,7 +43,7 @@ def plant_population
 
   def trait_descriptors
     tds = object.content.step02.trait_descriptor_list.try(:select, &:present?) || []
-    numeric_tds = tds.select { |td| td.to_s.match /\A\d+\z/ }
+    numeric_tds = tds.select { |td| td.to_s.match(/\A\d+\z/) }
```
Nuanda: I understand `numeric_tds` refer in fact to `existing` TDs that are already present in the DB. If this is indeed the case, can we change this naming a bit? Since we use 'new_trait_de....; elsewhere, can we call the existing ones `old`?

BTW what happens to that match if one calls her new TD with a number?

app/forms/submissions/trial/step03_content_form.rb
kammerer: It was supposed to protect against empty string, but I think it is not possible right now to pass empty string here. Besides, this method is not even called due to the typo in the name. I will remove it.

app/decorators/plant_trial_submission_decorator.rb
kammerer: I used `existing_` before as it seemed more appropriate than `old_` but I can change it to `old_` for the sake of consistency.

It seems to me (from checking the code) that naming a TD with a number will result in:

- either ignoring that TD altogether (if no TD with given id exists),
- or assigning existing TD instead of using TD data provided.
  
Either way, it is a bug.

Nuanda: Yes, it seems we have `old_` in two places and `existing_` in two others. So please change to `numeric` to `existing` and also try to replace the `old_` occurrence in `submission.rb` (the other one is in specs, you may ignore that one).

Currently one is not able to call a new TD with a number, since I closed the vocabulary (it's only Trait.name values now). However, I can easily imagine us allowing a free text version, since the TO ontology is rather limiting for our users. Therefore I asked about that. Are existing `tds` stored in the array as integers?

kammerer: I believe that params passed from for are always seen as strings and it
stays like that when they are stored in submission's content.

2016-06-16 19:20 GMT+02:00 Tomasz Gubała <notifications@github.com>:

> In app/decorators/plant_trial_submission_decorator.rb
> <https://github.com/eSpectrum-IT/brassica/pull/546#discussion_r67386923>:
>
> > @@ -43,7 +43,7 @@ def plant_population
> >
> >    def trait_descriptors
> >      tds = object.content.step02.trait_descriptor_list.try(:select, &:present?) || []
> > -    numeric_tds = tds.select { |td| td.to_s.match /\A\d+\z/ }
> > +    numeric_tds = tds.select { |td| td.to_s.match(/\A\d+\z/) }
>
> Yes, it seems we have old_ in two places and existing_ in two others. So
> please change to numeric to existing and also try to replace the old_
> occurrence in submission.rb (the other one is in specs, you may ignore
> that one).
>
> Currently one is not able to call a new TD with a number, since I closed
> the vocabulary (it's only Trait.name values now). However, I can easily
> imagine us allowing a free text version, since the TO ontology is rather
> limiting for our users. Therefore I asked about that. Are existing tds
> stored in the array as integers?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/eSpectrum-IT/brassica/pull/546/files/6e425ba70db7c5a52ebf2ed70395ee5b74d3bf9c#r67386923>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAA2ySvGxpI6DzfbLL0FiK456QvZfIo3ks5qMYXrgaJpZM4I2hPI>
> .
>



-- 
Tomasz Szymczyszyn

Nuanda: Ha, no free lunch, then... Let's leave it as it is, it's not worth delaying that work because of that.

Nuanda:
No more comments wrt the code. I'll be testing that manually today, please add specs in the meantime.

Nuanda:
Changing design factor names doesn't trigger the "not saved" alarm when clicking "BACK" (changing tech rep number does).

Nuanda:
When I have >1 DF names chosen (from the predefined ones) and I remove them all at once, by clicking the "x" button by the first name, and then I pick the first DF name again, the free text option in the second DF combo form is disabled, though it should not be. One needs to add the 2nd DF name from predefined ones, and then remove it with the 'x' button, and the free text field is enabled again.

A similar, though a bit different, thing happens the other way around. I add two free text DF names, then I remove both of them (at once) by clicking the 'x' icon by the first name. When I type in the first DF name in the free text, the second DF name is recreated with the old value I used before. This is not bad in itself, I can leave it, or remove it and then use another value or pick a predefined DF name. However, when I (after removing both DF names with the first name's 'x' button) pick a predefined value for the 1st DF name, the second DF form has the predefined select box disabled yet _no_ value in the free text field (it has the placeholder text inside it). I can still click the 'x' button making the select box enabled, but it's kind of unnatural to be required to do that since the entire combo widget looks like a blank one.

Nuanda:
@kammerer I have added an additional 'warning' section to the parser output, for permanent (i.e. not validation-related) errors and problems. I use that to remind the user to submit missing plant lines (so when the finalizer rollbacks submission to step 4 the user knows what's wrong with the scoring).

kammerer:
Fixed JS issues that you mentioned.

kammerer:
@Nuanda I updated submission specs and merged in technical_replicates branch.

Nuanda:
Do you plan to add anything else or it's ready for merge?

kammerer:
Ready for merge.

PR #553, opened 2016-06-16
Extended design factor model

### LINKED ISSUE #545: Promote DesignFactors as a writable resource (needed for raw data submission via API)

### LINKED ISSUE #549: Show design factor in PSU data table

### LINKED ISSUE #550: Refactor DF.design_factor_X columns to an array datatype column
Required for #543 and for #551.

IMPORTANT: don't forget to update the Plant Trial Finalizer, when merging #353 in.

Nuanda:
* turns design_factor_x columns to design_factors array column, fixes #550 
 * makes DF API-enabled, fixes #545 
 * shows DF.design_factors in PSU data table, fixes #549 
 * also, by the way, makes PSUs searchable (should be done some time ago), also tests searching PSU with a single desgin_factors item.


Nuanda:
@nowakowski Piotr, please review migrations.

@kammerer Tomasz, please review the API-related and ES-related stuff, esp. the `CreateParams` changes.

db/migrate/20160615154337_move_design_factor_columns_to_array.rb
```diff
@@ -0,0 +1,39 @@
...
+    add_column :design_factors, :design_factor_1, :string, null: true
+    add_column :design_factors, :design_factor_2, :string, null: true
+    add_column :design_factors, :design_factor_3, :string, null: true
+    add_column :design_factors, :design_factor_4, :string, null: true
+    add_column :design_factors, :design_factor_5, :string, null: true
+
+    design_factors = DesignFactor.all.pluck(:id, :design_factors)
+    design_factors.each do |id, factor_array|
+      values_string = factor_array.map.with_index do |value, i|
+        "design_factor_#{i + 1} = '#{value}'"
```
nowakowski: This is vulnerable to SQL injection and will also cause problems with special characters in factor names. I've written a helper method which helps with this issue - see `20160610151911_edit_factors.rb` for an example on how to use it (should be self-explanatory).

nowakowski:
One issue identified in migration file; otherwise :+1: from me.

db/migrate/20160615154337_move_design_factor_columns_to_array.rb
Nuanda: Fixed.

app/models/design_factor.rb
```diff
@@ -1,6 +1,13 @@
...
+  belongs_to :user
+
+  after_update { plant_scoring_units.each(&:touch) }
+  before_destroy { plant_scoring_units.each(&:touch) }
+
   has_many :plant_scoring_units
 
+  validates :design_factors, presence: true
+
   validates :design_factor_name,
```
kammerer: Out of curiosity, what is `design_factor_name` actually?

config/locales/api.en.yml
```diff
@@ -1217,6 +1216,64 @@ en:
...
+          name: date_entered
+        -
+          name: data_owned_by
+        -
+          name: data_provenance
+        -
+          name: confirmed_by_whom
+        -
+          name: design_factors
+          format: array
```
kammerer: Before, we denoted arrays as "int[]" (only integer arrays were in use). So, I would propose to either stick to this convention, or maybe change "int[]" to "array" too?

kammerer:
:+1: I added one more test case and a minor remark regarding docs.

app/models/design_factor.rb
Nuanda: An old CS column. Not really useful in this context. In CS times they used names as FKeys (not even real FKeys, just values), so they needed (sometimes artificial) name column everywhere.

Nuanda: Sorry, I forgot I've done that already in the other PR. Merging this one, then.

PR #554, opened 2016-06-16
Bring the Cropstore dump used in rake app:bootstrap up to date

### LINKED ISSUE #547: "Squash" current migrations by providing another sql dump for loading
As we have discussed yesterday, don't forget about the rake tasks (you have identified two) that should be invoked on the DB, before doing the dump.

nowakowski:
Fixes #547


Nuanda:
I can't check the data, obviously, but I run the bootstrap process and then I run all specs, so the schema should be up to date. :+1:

PR #559, opened 2016-06-18
PlantAccession and QTLJob models searchable

### LINKED ISSUE #552: Make PlantAccessions and QTLJobs searchable
Since they are shown in the browse section, they should be ES-searchable.

### LINKED ISSUE #557: Update API docs 'Q' markings
There might be some parameters which are now allowed for searching (fetching) though the API, but which are not yet denoted with a 'Q' in the docs.

Nuanda:
Fixes #552 

Also fixes #557


kammerer:
:+1:

PR #563, opened 2016-06-23
Fixes in search

Nuanda:
Several updates to ES indexing, some small fixes.

### DIFF
--- a/app/models/map_locus_hit.rb
+++ b/app/models/map_locus_hit.rb
@@ -74,7 +74,7 @@ class MapLocusHit < ActiveRecord::Base
     indexes :atg_hit_seq_source
     indexes :bac_hit_seq_id
     indexes :bac_hit_seq_source
-    indexes :bac_hit_nameindexes
+    indexes :bac_hit_name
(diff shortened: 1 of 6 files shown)


Nuanda:
@kammerer Tomasz, I made a few changes to ES indices - some because the structure of browse section data tables changed (and it only makes sense to search through what is visible to the user), and some are related to the issues (?) detected this morning.

Please review.

One more thing remains - TS table is not indexed right now. It's a lot of data yet I think we should do that, at least give it a try (and see what kind of delays/mem consumption we have).

kammerer:
What ES version are you using?

kammerer:
Looks good :+1:

Nuanda:
2.3.3

kammerer:
@Nuanda We use 1.5 in production...

Nuanda:
I know, my OS upgrade did it to me (I use system-provided ES). So I deployed and tested it a second ago, on production, and it looks like MP.map_position works as intended (only full matches are honored).

PR #569, opened 2016-06-24
555 ignoring pl pv for existing pa

### LINKED ISSUE #555: Trial submission parser should not require PL/PV values for known PAs
As we don't really need that value in the finalization process.

Nuanda:
Fixes #555


kammerer:
:+1:

PR #577, opened 2016-07-09
Assigns missing originating organisation for PAs

### LINKED ISSUE #560: Solve problem of current no-organisation accessions in BIP
Let users utilize these accessions as well in their submissions.

### LINKED ISSUE #568: Originating organisation for 288 Plant Accessions
We have 288 Plant Accessions which do not have an originating organisation. Since we started to require that value (OO) for new accession, it would be even better if we could set OO for all present values of PA.

All these 288 PAs have the following `plant_accession` column values:  "rres_GK090855", "rres_GT080921", "rres_GT090261", "rres_GK090235", and so on. All of them are either 'rres_GK' or 'rres_GT' followed with a number.

According to CS data all of them were "entered_by" Pierre Carion (BBSRC), "confirmed by" Jackie Barker (BBSRC) and, as "accession_originator" (?) they have either Jackie Barker or Graham Teakle (Warwick).

Annemarie, are you able to assign an originating organisation for these PAs?

Nuanda:
Fixes #568 

Fixes #560 

In consequence, also kind of fixes #560 - there are no longer null-oo PAs in the DB and there should be none in the future, since we now require that column.

I stick with just the mnemonic RRES since this is how they call Rothamsted in other parts of CS data.

### DIFF
--- a/db/migrate/20160709082305_assign_missing_originating_organisation.rb
+++ b/db/migrate/20160709082305_assign_missing_originating_organisation.rb
@@ -1,0 +1,9 @@
+class AssignMissingOriginatingOrganisation < ActiveRecord::Migration
+  def up
+    execute "UPDATE plant_accessions SET originating_organisation = 'RRES' WHERE originating_organisation IS NULL"
+  end
+
+  def down
+    execute "UPDATE plant_accessions SET originating_organisation = NULL WHERE originating_organisation = 'RRES'"
+  end
+end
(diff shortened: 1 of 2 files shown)


Nuanda:
@kammerer This Travis error - could that be connected with the repository switch? It looks like the ENV variables for testing were retained from the old build setup...

kammerer:
:+1:

PR #584, opened 2016-07-10
Update deps

kammerer:
Update dependencies with known security vulnerabilities. I've more-or-less tested the application manually, all specs pass to.


Nuanda:
Will capistrano fetch the new gem versions on deploy?

kammerer:
Yes.

Nuanda:
Updated gems locally and run all specs, so I guess this is :+1:

Nuanda:
@kammerer I think I get a lot of DEPRECATION warnings after merging this in. Also Travis does. Do you know what is the reason of that?

kammerer:
Probably updated RoR introduced those.

PR #592, opened 2016-07-13
Api client example

### LINKED ISSUE #394: Prepare script to load RIPR data through BIP REST API
Take care to extensively document the script as it is a kind of deliverable for the users - they will use it to prepare their own versions for uploading further RIPR (or other project) data to BIP.

Nuanda:
Fixes #394 

Tried to use only stdlib ruby. Provided fairly simple code extraction to functions, but no OO (not to make it overcomplicated). This is an example, _not_ a general-purpose BIP client. It should show how one _could_ build a client to submit a given Plant Trial data.

Run with the data file from the same directory.


Nuanda:
I realized I need to add the API docs section describing this...

Nuanda:
@kammerer Do you think I could test it (running the client in external ruby process) with capybara? I mean - I need a standalone server on localhost, running the test env, which I can explicitly call (using a URL) in the client code.

Nuanda:
Since this example is badly needed, I'll leave creating a spec for that for another issue. So please review what we have now.

Nuanda:
@nowakowski Piotr,  could you have a look at the piece of docs I wrote here, and fix any grammar/style problems?

nowakowski:
@Nuanda Done.

Nuanda:
@kammerer Can you have a look at it?

kammerer:
:+1:

Issue #494, opened 2016-05-03
Population Submission- content and compulsory fields


teatree1212:
With regard to Subject 6 in issue #488 (restated here:
 when submitting a Population, e.g. 100+ single Lines may need to be submitted. This means for now in the wizard submission, that the user needs to manually create all these 100+ lines and add a taxonomy term to them. A few issues arise from here.)

Issue 1:
1: For a crossing population, a taxonomy term does not need to be made compulsory for each line submitted, as they are derived from their parents.
2: For a mutagenised population, a taxonomy term does not need to be made compulsory for each line submitted, as the mutagenised seeds are all from one parent.
3: For a diversity foundation set, the taxonomy is different, and it hence needs to be a compulsory field for each line or cultivar alike.
4: For a diversity fixed foundation set, the taxonomy is different, and it hence needs to be a compulsory field for each line or cultivar alike.

On that note, I am  sharing the Dictionary, where I am working on a defition for "population" that includes all these different forms of genetic material you include in it in the [BIP:](https://docs.google.com/document/d/1t0WUiPJ-ggNaayDBkutfipwtfspYsoZBw7QGbMhuD1E/edit?usp=sharing)

Issue 2:
A submission of lines can be up to over hundred lines with different numbers (1 and 2), and cultivar names (3-4). It would be better do submit a list in .csv or even just .txt format. Necessary columns would be line name, taxonomic term, genetic status, cultivar name, of which line name and taxonomic term definitely should be made compulsory.

Nuanda:
@teatree1212 Annemarie, can we put this issue on hold, until we sort out how to rebuild the Trial submission process, so then we can come back to this one? I mean - there are a lot of discussion points that remain opened at the moment... But if you think this one is required _before_ we deal with the trial submission redesign...

teatree1212:
of course. I thought I just mention this.

Nuanda:
Currently, we have the following `population types` in BIP:
 "F2",
 "DH segregating",
 "Integrated",
 "Genetic resource collection",
 "Diversity core collection",
 "Diversity foundation",
 "Substitution lines",
 "Experimental collection",
 nil,
 "Variety resource collection",
 "Recombinant inbred",
 "F3 pooled",
 "F3"

(nil means there are populations that never had any type assigned). Can you tell me how do these values map to the bigger four "type groups" that you've listed? I understand that several terms from this list map to a single term from your list.

Commenting on the four alternatives that you've given:
[...]

PR #601, opened 2016-07-15
Population submission example

### LINKED ISSUE #632: Create API client which removes PP-related data from BIP
Script which would remove:

  - a given PlantPopulation
  - all PlantLines related to this PP
  - all PlantAccessions related to these PLs
  - all PlantVarieties related to these PLs.

Nuanda:
We try to submit some PVs, some PLs, some PAs and a PP.

Also: Fixes #632


Nuanda:
@teatree1212 This is a pull request. When you click the "Files changed" above you'll see what is the current state. For now I did not change anything.

Nuanda:
Now, what I did I run the script with the following command (first navigating to the directory where it is located):

```
ruby population_submission.rb merged_with_UoYHybrids.csv my_api_key
```

and it successfully created the population (see https://bip.tgac.ac.uk/browse_data and filter with RIPR).

teatree1212:
yes, it did, I did that before, but removed the population afterwards

Nuanda:
All right  - so your api key and your ruby interpreter work ok. Now we need to work our way bottom - up. Do you want to submit this population, plant lines and plant varieties?

teatree1212:
it is a dummy sheet, not 100% correct, as I still need to confirm with people about how to name the accessions.

Nuanda:
`CSV.foreach(ARGV[0])` is the code that uses ruby std library to open the csv file, parse all lines and put them in ruby arrays. Each array (line) is called exactly once in the loop body (between `do` and `end`) and the array with data is called `row`. So, in ruby syntax, `row[0]` means the first element, `row[3]` means the 4th element etc.

teatree1212:
yes, this is why I defined PLANT_ACCESSIONS =0, you can see this further up in the script

Nuanda:
I have just committed a change where I check if the first element of the array equals 'Accession_name' - you will notice this is how you called the first header element. This way we will omit the header in the script (which is ok, since the header is rather for us humans, not for the script).

Nuanda:
Shouldn't VARIETY = 3, not 4?

teatree1212:
oh yes, I lost count ...^^ all of them need to be moved one forward

Nuanda:
So, for each row in the loop, we will call the `record_plant_variety` with the 2 values parsed from the CSV file (see the last commit).

teatree1212:
how come only 2? how is that shown in the code? I see you did this:
+  record_plant_variety(row[VARIETY], row[CROP_TYPE])

Nuanda:
No no - by two values I mean the variety name (1) and the crop type (2). In the function signature (the header) you define the so-called parameters - which you use in the function body (see lines 90 and 91). When you call that function (line 140) you assign "values" for that parameters.

Nuanda:
Now, we have a parsing problem with row Notthm_BnASSYST-144 - ruby claims the file is not UTF-8.

teatree1212:
thats weird, how is it different from the others..

Nuanda:
Can you paste the correct name for "G�lzower �lquell" here?

Nuanda:
(it might be that the encoding you used broke this particular name - though I'm not sure at the moment)

teatree1212:
i see the guelzower oelquell problem again..

teatree1212:
The variety name is screwed up

Nuanda:
BnASSYST-282 will probably have the same problem.

teatree1212:
I haven't run it since I pulled it after your changes, and I am getting this error message. What does that mean?  
This is my output : 
1. Creating experimental plant_population 
{"status"=>"500", "error"=>"Internal Server Error"}
population_submission.rb:61:in `create_record': undefined method `[]' for nil:NilClass (NoMethodError)
	from population_submission.rb:75:in `<main>'

Nuanda:
Thanks. Please do the same for 195 and 282

teatree1212:
yepp

Nuanda:
It's Piotr doing the maintenance - sorry for that.

Nuanda:
But tell me if you understand my last change and how I call the function in the loop?

teatree1212:
i understand what you want to do. And i see what you did:
You changed the names from row[Variety] to plant_variety_name. By then calling record_plant_variety(value1, value 2), you import these. How does the function know that the header VARIETY goes to plant_variety_name:plant_variety_name and CROP_TYPE to crop_type: crop_type?

teatree1212:
Is it just the order?

Nuanda:
That's a trick we do using our knowledge of the CSV file header. Yes, we _know_ where is value is located (which column index) so we use the indices. So, `VARIETY` is simple a `constant` which equals 3 throughout the entire script execution (you could simply put the digit `3` instead `VARIETY` in the code and it would work as well).

teatree1212:
okay, so you run record_plant_variety(3, 4), but how does
def record_plant_variety(plant_variety_name, crop_type) know that 3 is plant_variety_name and 4 is crop_type?

Nuanda:
We do something more than that. `row[VARIETY]`, or, as we know, `row[3]` means - take the 4th value (remember ruby arrays start with 0, not 1) from the `row` array. And we know `row` equals the current row from the CSV file. Therefore, for e.g. row number 7, `row[VARIETY]` should equal "Milena".

Nuanda:
Likewise, `row[CROP_TYPE]` should equal "Modern winter OSR".

Nuanda:
Now, having these two values, call the `record_plant_variety` function, passing them as the two parameters. The function signature (line 85) defines, that the first parameter passed to that function will be available inside that function body (between `def` and `end`) under the name `plant_variety_name`, and the second parameter will be called `crop_type`.

So in this case - yes, it's the ordering of parameters that matters.

teatree1212:
yes, because ruby reads .csv s in rows I have read. Not very intuitive but good for that function

Nuanda:
Now, to the id assignment. The execution of the `record_plant_variety` function in line 140 should yield us the id of the plant_variety. Let me do further changes to the script.

Nuanda:
https://github.com/TGAC/brassica/pull/601/commits/7f5f113dba8ebd495ae0700b006e240b80987ab3

Nuanda:
So, I changed that `record_plant_variety` execution in a way that now we capture whatever it returns (and it returns the ID of the plant variety inside BIP) under a new variable, called `plant_variety_id`.

And then I pass the value of that variable to `record_plant_line` execution, just one line below.

Nuanda:
I have run this and you can see

https://bip.tgac.ac.uk/data_tables?model=plant_lines

that plant lines are registered and they have plant varieties associated with them.

teatree1212:
okay, now you are calling the function and assigning the output a variable ( correct vocabulary?)

teatree1212:
and you are also calling another function in which you use that variable. How does the record_plant_line function know which plant_variety_id is which? does it again depend on the order ( which here is consistent)
?

teatree1212:
But I thought in the variable(?) plant_variety_id, there would be So much content,.. but actually.. looking at the function create_record, all it actually records and therefor will be in that variable plant_line_id  may be the respective class_name id ??

Nuanda:
It kind of doesn't know. It doesn't need to. See - both functions are executed in the context of the same CSV row. So _we_ (the programmers) know, that variety from row 7 of CSV should be related to plant line from row 7 of CSV, right?

The `record_plant_line` just get a number and it does not know what PV it represents. It even does not know if the represents _any_ PV (though in that case the execution would break, since the BIP server DO check, if we pass existing PV identifier as `plant_variety_id`). It just sends it to BIP as the value for the `plant_variety_id` parameter.

However, when we call that function we know that this variable holds id for the PV that we have just registered, and we know it is the PV from the same CSV row, so it is the correct id.

Nuanda:
Yes, `create_record` (see its last line) returns only the 'id'.

Nuanda:
Also (see line 94) if the PV in question already exists in BIP, this line will extract its 'id' and return that. So either way (existing PV/newly created PV) the function should return the id (which is a single integer number).

teatree1212:
yes, I see! So now I suppose the population submission works, the plant_varieties, too, the Plant lines should, as well, because we are also calling that function, so we are left with the accessions. Can I try and  suggest changes first

Nuanda:
(I will remove our added content from the DB so we can do one more thing)

Nuanda:
One minute, let me do the last change.

teatree1212:
ok

Nuanda:
Because we still need to associate the lines with the population. I understand we should use the `plant_population_list` table, which associates PLs with PPs.

Nuanda:
https://github.com/TGAC/brassica/pull/601/commits/64976103b74c9fa734d9c7dfbe70a1e56d1418d3

I added the `plant_population_id` variable to remember the id of the plant population we created (note we have single PP for the entire submission). I also recorded the id of a PL as `plant_line_id` (note we have different line for each CSV row).

Nuanda:
And I added a new `associate_line_with_population` function which created a record in the `plant_population_list` table - and that record associates a given PL with the PP. There is one such record per PL we create.

Now, when you go to https://bip.tgac.ac.uk/browse_data and filter by RIPR, you'll see our PP. There, you can click Related and this will get you to the list of PLs.

teatree1212:
okay, got it. The principle is always the same really. You create functions that can draw from the csv file ( not in case of the population though) and can send the contant away via the api.  Then you call the functions.

Nuanda:
Yes. Basically you don't really need functions - you could put that code directly inside the `CSV.foreach` loop, but having functions make the code more readable and manageable.

Nuanda:
Now, did you notice there are only 133 PLs registered while we have 393 (or so) rows in the CSV?

teatree1212:
I was wondering how the script knows which csv to draw from if there are multiple ones?

Nuanda:
It seems we have some data problem.

Nuanda:
BIP, for good reasons, exclude the possibility that we create two `plant_population_list` records with exactly the same pair of PP and PL. And this happens with `U.Notthm_BnASSYST-151` in our data. You will notice the duplication there, in the CSV. So my script exited informing me that such a record already exists (correctly). Now we can either make it ignore this, and, probably better, we should correct the CSV file.

Nuanda:
Are you able to correct that PL?

teatree1212:
it is a dublication in the plant line Librador. The reason for that is that in one case the SRR identifier is different. this plant has been sequenced twice and therefore there are two sequence identifiers associated with it. I thought I just dublicate everything else but now I see that it didn't like it. Maybe we cannot only test for the presence of the variety name, but - in this case- for the similarity between SRA_identifiers

teatree1212:
there are a few more such cases in this dataset. Maybe, it is better to have it as one row, but have both identifiers ";" separated in that location

Nuanda:
Hmm, this is surprising - I didn't know it might be the case. BIP Plant Line model put so-called _uniqueness constraint_ on the `plant_line_name` column in the `plant_lines` DB table. This is in order to  prevent people from submitting the same PLs over and over again (if it's the same `plant_line_name`, BIP block the creation of the record in the DB).

Nuanda:
Now I understand this is a bit too simplistic - there could be same line names representing (genetically) different organisms? Or maybe we should have them called differently, then?

teatree1212:
oh no, it  is the same line. but the material from that same line has been sequenced twice. It is a bit like a replicate measurement.

teatree1212:
So I will edit these duplicate as I suggested and have the two identifiers in the same entry. This makes more sense. I just didn't know how the submission works and hoped it would just add this one different element to the existing record.

Nuanda:
If we followed "the DB wudan path" ;) we would create another table/model/resource, called PlantLineSequence, and have PlantLines relate to 0+ records in this new table. So every PL can have 0+ sequences registered.

But I'm not sure it's worth it at the moment. Would registering all sequence identifiers as an array be acceptable?

Nuanda:
For example, see Design Factor create API here: https://bip.tgac.ac.uk/api_documentation#design-factor

Nuanda:
This, however, requires changes in BIP. For now, could you try to submit that as a single PL, with the sequence_identifier value like that: "SRR3138274,SRR3134144"?

teatree1212:
yes, I thought just having that information stored in the plant_accessions.comments column for now will have the information associated with the right things and we can think what we want to do with it until later.

teatree1212:
I have now changed them and separated them by ;  thinking that otherwise it would screw up the  csv, was that ok?

Nuanda:
Wrong paste in the 'Marnoo' row?

public/population_submission/merged_with_UoYHybrids.csv
```diff
@@ -236,8 +233,7 @@ U.Notthm_BnASSYST-274,BnASSYST-274, SRR3134242,Ceska Krajova,Spring OSR,Nottingh
 U.Notthm_BnASSYST-275,BnASSYST-275, SRR3134243,Duplo,Spring OSR,Nottingham,Brassica Napus
 U.Notthm_BnASSYST-276,BnASSYST-276, SRR3134244,Janetzkis Sommerraps,Spring OSR,Nottingham,Brassica Napus
 U.Notthm_BnASSYST-277,BnASSYST-277, SRR3134245,Line,Spring OSR,Nottingham,Brassica Napus
-U.Notthm_BnASSYST-278,BnASSYST-278, SRR3138275,Marnoo,Spring OSR,Nottingham,Brassica Napus
-U.Notthm_BnASSYST-278,BnASSYST-278, SRR3134246,Marnoo,Spring OSR,Nottingham,Brassica Napus
+U.Notthm_BnASSYST-278,BnASSYST-278, SRR3138275,Marnoo;SRR3134246,Spring OSR,Nottingham,Brassica Napus
```
Nuanda: @teatree1212 Here :).

teatree1212:
oh.. S: one sec

teatree1212:
I will have to go home and have some food after this but will be back afterwards, if you are still working ? Maybe this script needs a warning when wanting to submit the same plant_line again. They should be unique.

Nuanda:
It does warn - it returns the 'Has already been taken' error.

Nuanda:
This is possible to program, of course. It would simply need to be more complex.

teatree1212:
nooo, it didn't save!

teatree1212:
oh yes it should, but  with  git log -p it shows the same mistake still.

Nuanda:
It should be correct, now.

teatree1212:
ohh. okay, it shows changes from multiple commits .. sorry, I have learned how it works in theory, but doing it is still different and my eyes are not familiar with what they see.

Nuanda:
Sure, go ahead and have something to eat. I will probably need to stop soon as well - but I'll add the sequence recording bit.

Nuanda:
I added recording the sequence identifiers. Now it works ok - there are 389 PLs associated with the PP, and each PL has a related PV. SRRs link to NCBI. The only thing is that my javascript code does not expect "SRR...;SRR..." multiple sequence ids (so the URL is wrong) but that could be mended in the future.

teatree1212:
Oh cool, that's great! The SSR identifiers need to however be associated with the Accessions and not the lines.  they should ideally end up in plant_accessions.comments. And then we have to decide whether to establish a Sequence column there, too.

teatree1212:
I actually think that you need to use this url for the query:
http://trace.ncbi.nlm.nih.gov/Traces/sra/?run=SRR3134012

The point is that at some point we want to  retrieve the sequences ( genotypic) material and compare it with the trait_scores that are in the BIP. your query goes to the Experiment-description, but not the sequence itself.

Nuanda:
Oh, please make an issue about that.

Why accessions? Does it mean we have displaced the sequence_identifier column and it should be inside PA, not PL?

Nuanda:
(I have removed submitted data from BIP, so you are free to run the script and record things using your api key)

teatree1212:
ahem, could you remove it again? I made a mistake pretty far down..

Nuanda:
What do you want me to remove? All PLs, PVs and PPs with your user_id?

teatree1212:
yes please

Nuanda:
Done. I'll try to make a simple example (next week maybe) how to delete your records through API.

teatree1212:
I know how to do that for single entries, but for multiple accessions for example, I don't know. I suppose it works by the same principle as this script?

Nuanda:
Yes, you would normally iterate over accessions and delete them.

teatree1212:
could you repeat the deletion again please S: ?

Nuanda:
I did.

Nuanda:
BTW do you use MacOS or Linux by any chance?

teatree1212:
MaxOS

Nuanda:
Because if you do you could try to setup your own BIP development environment. It would allow you to test things against your own BIP instance before you run them against the production server.

teatree1212:
can you please repeat the deletion? I can't wait to learn how to do it myself.

teatree1212:
yes, that sounds good.

Nuanda:
Done. Setup of a full dev env it, however, a pretty complex stuff - and we do not use Macs in our company - so this might not be so easy...

teatree1212:
I seem to remember we had touched this area before in a conversation. Maybe I can just create a linux environment on my Computer ?

teatree1212:
I think it is working now, finally! yesss, thank-you for your help, that was fun!

Nuanda:
Is there any chance that you can ask Carlos (I think it was him) to help you setup the environment on your MacOS?

teatree1212:
yes, i will ask him.

teatree1212:
@Nuanda Can you delete my submission from last Friday night? Now that it works, I will identify the proper content for accession and line in this particular submission. Those are the only two columns that may not have proper content right now. I will use the script on the 4th when we submit those people's data.  You said you were writing an explanation on how to use the corresponding trait submission script, I was going to that with this one, too. if you already wrote yours, can I have a look so that I follow the same patterns as you? or were the comments in your script the explanations ?

Nuanda:
All PVs, PLs, PAs and PPs assigner to your user id should be removed. hasn't check any other class.

I wrote a short API Documentation section describing the client script example, you can see it here: https://github.com/TGAC/brassica/pull/592/files. However, I don't think that doing the same for this new population submission script is worth your time - since it basically uses the same techniques, so it will not introduce much more new information, tutorial-wise.

public/population_submission/population_submission.rb
```diff
@@ -136,8 +143,12 @@ def record_plant_accessions(plant_accession, accession_originator,plant_variety_
 
 
 CSV.foreach(ARGV[0]) do |row|
-  next if row[0]== 'sample_name'
+  next if row[0]== 'Accession_name' # omit the header
   puts "  * processing Accession  #{row[ACCESSION_NAME]}"
+  plant_variety_id = record_plant_variety(row[VARIETY], row[CROP_TYPE])
+  plant_line_id = record_plant_line(row[LINE_NAME], plant_variety_id)
+  associate_line_with_population(plant_line_id, plant_population_id)
+  record_plant_accessions(row[ACCESSION_NAME],row[ACCESSION_SOURCE],plannt_line_id,row[SRA_IDENTIFIER])
```
Nuanda: I think there is a typo in this line.

public/population_submission/population_submission.rb
```diff
@@ -72,57 +72,64 @@ def test_error(response)
...
         request = Net::HTTP::Get.new("/api/v1/plant_accessions?plant_accession[query][plant_accession]=#{plant_accession}", @headers)
         response = call_bip request
         plant_accession_id = if response['meta']['total_count'] == 0
-                      plant_variety_id = record_plant_variety(row[VARIETY], row[CROP_TYPE])
                         create_record('plant_accession',
-                            plant_accession: row[ACCESSION_NAME],
-                            originating_organisation: row[ACCESSION_SOURCE],
+                            plant_accession: plant_accession,
+                            originating_organisation: originating_organisation,
                             plant_variety_id: plant_variety_id,
```
Nuanda: You probably don't want to do this. I think it is enough that a PV is related to a PL, since that PL is further related to a PV.

teatree1212:
Okay i won't write anything then.

Oh, yes I changed that typo, but didn't push the changes. 

did you mean PA instead of you last PV in you latest comment?

teatree1212:
I also changed the description of the tasks this Ruby client does, change it if you find better descriptions

Nuanda:
I removed both all PAs and all PVs you have created in the production DB.

teatree1212:
thanks.

Nuanda:
I have adjusted some comments and outputs so they more precisely reflect what is really happening during the script execution.

teatree1212:
maybe, in the future,expansion of the client would include the loading of the taxonomic term ( =species)that needs to be added to the single plant_line ( discussion in issue #494 ) . In the first submission to be happening soon, only brassica napus lines will be submitted, which makes it easy to just get the identifier.

Nuanda:
This is a question about vocabulary maintenance. The same thing is about adding new Plant Parts or Traits. Should it be freely allowed for the user? If so, we risk getting a lot of redundancy (it's faster to add your Trait instead of searching for an existing one and deciding if it fits or not). Otherwise, if we block that possibility and make people ask us to add new Traits or Taxonomy Terms ("controlled/closed vocabulary"), we waste our time.

What's your opinion?

teatree1212:
with regard to taxonomy id's: 
it would be good to have a control mechanism in place to ensure vocabulary maintenance.
- we could ask the user beforehand to identify the id and paste it into the respective columns ( this can contribute to  plant_lines falsely uploaded with wrong species tags). 
- the user could use species terms for the submission and while the client is running and it encounters a problem by not being able to assign a taxon_id to the species name that is supposed to be encountered ( e.g. spelling mistake, or the name is not written in the same way as in our db), we ask the user to input the correct taxonomy name for that particular plant_line and suggest what s/he could mean based on what was written in the field. After the user has manually corrected this input, the client keeps running until a similar problem is encountered. 
It would be better to the automatically correct all similar spelling or wrongly-written taxonomy terms that were previously manually corrected, so that the client only reports back with a taxonomy spelling issue when there is a new taxonomy term discovered that is ambiguous.

About Plant Parts
we could suggest plant_ part names from other ontologies as it is done now, but if one it is not in there we could ask " Is your Plant_part not in the list? - add it here"

With traits, I agree that it will be hard to enforce good practice. But in a different discussion we can talk about an approach where the user will be guided through all options step-by step, I have an idea but you have to tell me whether it is realistic and feasible to set up.

kammerer:
@Nuanda It would be good to verify that the data is correctly utf-8 encoded before performing any submissions. Otherwise we need to purge partial submissions from the database of something goes wrong.

Nuanda:
Do you mean the `CharacterNotInRepertoire` PG error?

kammerer:
That too. Obviously it is not possible, to prevent all encoding-related errors...

teatree1212:
this is the structure of the script we used for the submission. I altered the population sheet because some of the line names would have to be altered for rights reasons. I just cut out most of the content, so this is not a problem any more. That way I also got rid of names that can lead to formatting issues. If you prefer having a few in there, then I will add some again.

Nuanda:
@teatree1212 If the current script is to your liking, I'm ok with merging it to master.

Nuanda:
@teatree1212 I have created another useful cleanup script, see 'public/client_example/example_cleanup.rb' file. It removes **all** PVs, PLs, PAs and PPs owned by the user (not just the ones related to a certain population). It might be useful.

Nuanda:
@teatree1212 The cleanup script which removes all PLs, PAs, PVs for a certain PP, along with this PP, is under preparation in the `public/population_submission/population_cleanup.rb` file. I'll let you know when it's more or less ready.

Nuanda:
Ok, the population_cleanup.rb script is more or less working, please try to test it on your data.

public/population_submission/population_submission.rb
```diff
@@ -60,83 +68,85 @@ def create_record(class_name, data)
...
-puts '4. Submitting plant_accessions'
+# Function that finds or submits plant_lines
+def record_plant_line(plant_line_name, plant_variety_id, comments)
+  request = Net::HTTP::Get.new("/api/v1/plant_lines?plant_line[query][plant_line_name]=#{URI.escape plant_line_name}", @headers)
+  response = call_bip request
+  if response['meta']['total_count'] == 0
+    create_record('plant_line',
+      plant_line_name: plant_line_name,
+      plant_variety_id: plant_variety_id,
+      comments: comments #SRA identifier
```
Nuanda: @teatree1212 Shouldn't we use PL's `sequence_identifier` for that?

kammerer:
@teatree1212 @Nuanda I added another script `public/client_example/csv_check.rb`. It can be used to test file encoding. It does not do anything if the file is wrongly encoded, the author needs to fix it.

Nuanda:
@kammerer thanks! Could you add some information about that to the docs section?

public/population_submission/population_submission.rb
teatree1212: yes, sequence_identifier is the appropriate location. Sorry even though you talked about it here before, I wasn't using it in the submission. please migrate the identifiers over from PL.comments to sequence_identifiers. I will alter the script accordingly.

teatree1212:
I tested your population_cleanup.rb script. it removed the plant lines associated with the submission (178) that were different from the ones associated with Thomas' api_key. However, the possibility exists that plant line and variety are the same, but the plant accession is different( as the seed material would have been generated in a different year). this is what line 2 in merged_with_UoYHybrids1.csv emulates. This entry did not get removed when applying your script.
screen shot 2016-08-18 at 12 05 17

teatree1212:
Could I expand on the delete_mine function and add another if statement about 
"&&  date >[yesterday]" 
, so as to make sure it only deletes things I have screwed up in the closer time frame?

Nuanda:
Possible filtering options are explained in https://bip.earlham.ac.uk/api_documentation#filtering-data doc section. You can combine `only_mine` with these (see e.g. line 84 in the `population_cleanup.rb` script).

In case some resource attribute is not available for filtering, you are able to filter it out on the client side. See, for instance, the `delete_mine` function from the `example_cleanup.rb` script. You can try something like (lines from 62+):

```
response[plural_name].each do |record|
  if record['date'] > (Time.now - 1.day)
    request = Net::HTTP::Delete.new("/api/v1/#{plural_name}/#{record['id']}", @headers)
    response = call_bip request
  else
    @untouchables << record['id'].to_i
  end
end
```

where `date` is the column name you want to filter by.

Nuanda:
@teatree1212 Regarding your previous comment - the script, when removing PLs, PVs and PAs, does not rely on names, just identifiers. So, when removing a PL, it will remove all PAs which are related to this PL, and it will remove the PV, if it is related to this PL (related == the foreign keys in table rows are set).

Also, nothing which does not belong to you (i.e. was submitted by, or API-created by, someone else) should not be removed. Furthermore, published records past the revocation period (2 weeks) will not be removed either.

teatree1212:
yes,I understand that.
But in "my" case, I don't own the PL, but that was already submitted by Thomas earlier. I own the PA that is associated with it.
the script removes everything from a PL point of view, which is good and thorough, as PL and PA are associated with it.  
But I am guessing that if the PL already exists ( in this case submitted by Thomas), I might submit only the PA as a new submission  ( or this is what it seems to me and seems logical ). 
This means however, when the script tests for PL, it will not recognise that "behind" the PL submitted by Thomas is a PA submitted by me.

Nuanda:
I understand it now. You are correct, this will not be handled, since the script is rather conservative in this regard. Anyway, the scripts are just examples, the use cases could be do diverse that API users will need to modify them to their purpose anyway.

In this particular case you might want to try to remove the `only_mine` filter from PL GET call in line 56. This way, however, you will try to DELETE Thomas' PLs in line 68 (and, perhaps, also his PAs and PVs), which should result in ERRORs. Still, thus adjusted script might work and do what you want it to do.

teatree1212:
when Thomas was here, I was relying on Tomek to remove the full entry. How did he do it? I suppose that was him with the status as a database administrator, just wiping bits of /reloading the database ?
Earlier in this conversation we were trying so test submit,  and I had to rely on you to remove the entries. If me stitching the script and api documentation bits together will do that myself, then I will try. If there is anything else missing do to so myself, let me know.
Something unrelated: At the moment I am quite busy, so point me towards pressing issues that you would like me to comment on in particular.

Nuanda:
Yes, we do that directly on the production database - it is easier for us, as we know the internals of the application DB model.

Hopefully the cleanup scripts will work for the workshop attendees as well, so no direct administration intervention will be needed, but I should be here anyway that day to assist.

I am merging this.

PR #603, opened 2016-07-16
Api filter for owned

### LINKED ISSUE #598: API extension - filter by 'mine'
I think one interesting extension would be to filter records to get the ones that belong to me. Enabling `user_id` in permitted query params is not enough, as people do not know their `id`.

One possibility is to add a special purpose GET param, like `&only_mine=true`, for this.

Another is to follow the `query` convention, and enable `user` query param, to have something like this: `?plant_line[query][user]=me`.

Both ways the `ResourcesController.index` would need to treat this in a special way, transforming the param to a proper where constraint `user: @api_key.user`.

@kammerer Tomasz, what version would you prefer? Any other thoughts on that topic?

Nuanda:
Fixes #598


kammerer:
@Nuanda I moved some code from the controller to service classes. If it is OK for you, then feel free to merge. :+1:

PR #610, opened 2016-07-26
Does not require taxonomy term field in PP submission wizard 2nd step

### LINKED ISSUE #609: Taxonomy Term in PP submission 2nd step should be optional
Discussed in #494

Nuanda:
Fixes #609


kammerer:
:+1:

PR #614, opened 2016-08-01
Image url trait scores

### LINKED ISSUE #544: Data Submission_root data trial picture submission
a second issue about this root data, it has pictures and they would like to submit is somewhere. This is not BIPs problem at the moment, but it would be good to have some picture identifiers already submitted with the trial metadata.

Their pictures will be submitted to CyVerse, which will in the future be cross-linked to BIP.
As this is not the case yet, it would be however good to have some sort of
  **"picture name" column** in the TraitScores table, with which we can later identify each picture stored in CyVerse. So while the people submit their design_factors, trait_scores, they also submit picture_identifiers ( which will be multiple per sample, so maybe comma separated).

Does this sound like something that can reasonnably quickly be implemented?
Or if we used the API for the submission of this data, can we just agree on a column with text content, where we store this information to for now, which then can be repurposed?

Nuanda:
Fixes #544

### DIFF
--- a/app/assets/javascripts/data_grid_specific_columns.js.coffee
+++ b/app/assets/javascripts/data_grid_specific_columns.js.coffee
@@ -271,6 +271,10 @@ window.configs =
         render: (data, type, full, meta) ->
           modelIdUrl('trait_descriptors', data, full[full.length - 3])
       ,
+        targets: 'trait_scores_score_value_column'
+        render: (data, type, full, meta) ->
+          imageDataUrl(data)
+      ,
         targets: 'plant_scoring_units_scoring_unit_name_column'
         render: (data, type, full, meta) ->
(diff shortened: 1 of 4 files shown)


Nuanda:
@kammerer Tomasz, I detect the "special" path score value on the client side due to the fact that for one table view (the trail scoring view) I get only raw values from the server. I perform a simple escape there to properly show the URL.

kammerer:
:+1: 

The URL handling is not safe in general, since `,` is a valid character for an URL. I assume we can live with this for now since CyVerse image URLs are predictable and not affected by this.

teatree1212:
please have a look at issue #544 . I think the URLs need to be changed.

Nuanda:
ok, will see at it in a minute

Issue #616, opened 2016-08-01
Rebuilding index from scratch causes private data to be indexed


kammerer:
A new privacy aware task is needed to index data (instead of `rake environment elasticsearch:import:model CLASS='QtlJob' FORCE=y`)

PR #617, opened 2016-08-01
Privacy aware reindexing

### LINKED ISSUE #615: Deployment slow due to reindexing
Right now ES indices are reindexed on each deployment and that takes quite long time. However reindexing is not always necessary - it is needed only when we change the fields to be indexed. Two solutions come to my mind:

* enforce indexing on deployment with a special flag (skip it by default)
* enqueue indexing to be performed asynchronously after the deploy is completed 

@Nuanda WDYT?

kammerer:
Fixes #615 and #616 

After all, no special flag is needed for #615. Instead we will call `bin/cap production search:reindex:all` after performing deployment if rebuilding indices is needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: ruby
 rvm:
   - 2.2.1
 
+cache: bundler
+
 addons:
   postgresql: "9.3"
 
--- a/README.md
+++ b/README.md
@@ -1,15 +1,15 @@
 ## Brassica Information Portal
 
-The Brassica Information Portal (BIP) is a web repository for population and trait scoring information related to the Brassica breeding community. Advanced data submission capabilities and APIs enable users to store and publish their own study results in our Portal. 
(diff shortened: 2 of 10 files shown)


Nuanda:
Please update the README Deployment section (how one runs that task and when it is needed). Then: :+1:

kammerer:
I wanted to fix issues with specs before merging this, but it appears that it will require digging deeper into ES. The problem is that specs randomly fail due to ES indices not being ready to be queried. I don't however know what is the underlying reason for that (sleeping does not help much).

Nuanda:
No it doesn't, I tested that as well.

kammerer:
Perhaps this is an issue with my instance of ES which has some unassigned shards and stays "red", however Travis shows similar behaviour.

kammerer:
I fixed issues with my ES cluster and now specs pass consistently, so I think we can merge and fix travis later. Also, I enabled bundle caching on travis so that the build is a bit faster.

PR #631, opened 2016-08-09
Shows PVname in PA table when PA is related to a PL

### LINKED ISSUE #626: Show related PVName in PA table and in PA Zenodo export (if related though a PL)

Nuanda:
Fixes #626

### DIFF
--- a/app/models/plant_accession.rb
+++ b/app/models/plant_accession.rb
@@ -25,7 +25,6 @@ class PlantAccession < ActiveRecord::Base
 
   include Relatable
   include Filterable
-  include Pluckable
   include Searchable
   include Publishable
 
@@ -40,8 +39,7 @@ class PlantAccession < ActiveRecord::Base
         pv_subquery.as('plant_line_varieties').on { plant_line_varieties.id == plant_lines.plant_variety_id }.outer
     ]}
     query = (params && (params[:query] || params[:fetch])) ? filter(params, query) : query
-    query = query.
-      where(arel_table[:user_id].eq(uid).or(arel_table[:published].eq(true)))
+    query = query.where(arel_table[:user_id].eq(uid).or(arel_table[:published].eq(true)))
 
     query = join_counters(query, uid)
     query.pluck(*(table_columns + privacy_adjusted_count_columns + ref_columns))
--- a/app/models/plant_accession.rb
+++ b/app/models/plant_accession.rb
@@ -28,7 +28,24 @@ class PlantAccession < ActiveRecord::Base
   include Pluckable
   include Searchable
   include Publishable
-  include TableData
+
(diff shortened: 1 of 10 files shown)


app/models/plant_accession.rb
```diff
@@ -25,10 +25,25 @@ class PlantAccession < ActiveRecord::Base
...
+
+  def self.table_data(params = nil, uid = nil)
+    pv_subquery = PlantVariety.visible(uid)
+    pl_subquery = PlantLine.visible(uid)
+
+    query = all.
+      joins {[
+        pl_subquery.as('plant_lines').on { plant_lines.id == plant_accessions.plant_line_id }.outer,
+        pv_subquery.as('plant_varieties').on { plant_varieties.id == plant_accessions.plant_variety_id }.outer,
+        pv_subquery.as('plant_line_varieties').on { plant_line_varieties.id == plant_lines.plant_variety_id }.outer
```
nowakowski: Note: This could potentially be changed to an inner join if every PlantLine was linked to a PlantVariety -- however a look at the production DB indicates that this is not the case. I guess it's safer to leave it as is.

app/models/plant_accession.rb
```diff
@@ -41,7 +56,7 @@ def self.table_columns
     [
       'plant_accession',
       'plant_lines.plant_line_name',
-      'plant_varieties.plant_variety_name',
+      'CASE WHEN plant_varieties.id IS NULL THEN plant_line_varieties.plant_variety_name ELSE plant_varieties.plant_variety_name END AS plant_variety_name',
```
nowakowski: Note that this can still return NULL if the accession is linked to a PlantLine which is **not** linked to any PlantVariety. Perhaps that's acceptable - but is it?

spec/models/plant_accession_spec.rb
```diff
@@ -71,5 +90,17 @@
       expect(pa1).to be_valid
       expect(pa2).to be_valid
     end
+
+    it 'properly serializes pv.name' do
```
nowakowski: My spec essentially duplicates your spec (guess we both committed at the same time). I think one of those could safely be removed.

nowakowski:
I'm still getting 503s locally but I guess that's not related. Anyway, several inline comments for you to ponder - otherwise :+1:

app/models/plant_accession.rb
Nuanda: I know, and model/plant_accession_spec.rb should test that case (the example in line 16).

spec/models/plant_accession_spec.rb
Nuanda: Fixed.

app/models/plant_accession.rb
Nuanda: Yes.

PR #638, opened 2016-08-11
Zenodo bip backlink

### LINKED ISSUE #633: Try to submit BIP Trial/Population back-link along with Zenodo metadata
The idea is to have a URL that leads back from Zenodo deposition page directly to BIP. Even better - to the exact submitted object (either a PlantTrial or a PlantPopulation).

Idea sent by Thomas Alcock and @teatree1212.

Nuanda:
Fixes #633

### DIFF
--- a/app/decorators/submission_decorator.rb
+++ b/app/decorators/submission_decorator.rb
@@ -1,6 +1,19 @@
 class SubmissionDecorator < Draper::Decorator
   delegate_all
 
+  def self.decorate!(submission)
+    return submission if submission.is_a?(SubmissionDecorator)
+
+    case submission.submission_type
+    when "population"
+      PlantPopulationSubmissionDecorator.decorate(submission)
+    when "trial"
+      PlantTrialSubmissionDecorator.decorate(submission)
+    else
+      raise NotImplementedError
+    end
+  end
+
   def submission_type_tag
     h.content_tag(
       :span,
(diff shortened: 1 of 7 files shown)


Nuanda:
@kammerer To my surprise, something like that (see travis.yml) seems to work. I have repeated the build three times and it didn't fail :).

I will test again without the sleep.

app/decorators/submission_decorator.rb
```diff
@@ -33,4 +28,22 @@ def doi
...
+  def bip_link
+    if object.submitted_object
+      h.data_tables_url(self_url_params)
+    end
+  end
+
+  private
+
+  def self_url_params
+    nil unless object.submitted_object
```
kammerer: Missing `return`?

app/decorators/submission_decorator.rb
```diff
@@ -33,4 +28,22 @@ def doi
   def label
     raise Exception.new('Should be extended by subclasses')
   end
+
+  def bip_link
```
kammerer: It returns url, not link (like e.g. `link_to`), so `bip_url` (or `data_tables_url`, IMO better).

kammerer:
I added missing `return` and removed some repeated code.
You may consider my suggestion about method name. Besides that, it's OK :+1:

PR #645, opened 2016-08-11
Zenodo deposition tweaks

### LINKED ISSUE #630: Check if Windows newline are ok with the Zenodo contributors metadata parser
Possible bug in ZenodoDepositor line 92.

kammerer:
- Fixes #630 
- Simple refactoring of `ZenodoDepositor`
- Fail with error in case of broken deposition instead of showing a message to the user*
- A bunch of minor tweaks

[*] Actually, it may be OK to show the message, but we also should be notified


Nuanda:
Made some further changes. If you are ok with them, please merge to master.

PR #649, opened 2016-08-17
Touch cascade optimization

### LINKED ISSUE #625: Submission processing too slow
When I accepted submission with plant trial data provided by Thomas I left it as private. Still, it took a lot of time to finish the action. I have not measured it precisely, but probably more than a minute.

Nuanda:
Fixes #625 

Due to my benchmarks (ES-enabled, development env.), for finalizing "published" PT which created 50 PSUs and 200 TSes, this reduced the operation time from ~33 seconds to ~7.5 seconds. Finalized submission `publish` and `revoke` actions should have even greater speedup (up to 8x faster). Also destroying complex PTs were very slow and this speeds that up considerably.

The problem was that a lot of records were `touched` multiple times, while only a single touch was needed. This is a recognized flaw in AR (https://github.com/rails/rails/issues/18606) which was eventually removed in Rails 5.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,8 @@
 source 'https://rubygems.org'
 
 gem 'rails', '~> 4.2.0'
+gem 'activerecord-delay_touching'
+
 gem 'pg'
 
 gem 'puma'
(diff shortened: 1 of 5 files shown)


kammerer:
:+1:

PR #652, opened 2016-08-22
Mass-inserts TSes during PT submission finalization

### LINKED ISSUE #625: Submission processing too slow

Nuanda:
Another optimization in PT submission. This time we bulk-insert TS data for each PSU, and we update TD.tsc and PSU.tsc counters only once.

For 380 PSU and 22 TS per PSU (+8k TSes in total) - our current benchmark - it reduces published submission finalization >7 times (from over 7 min to 53 secs).

Fixes #625

### DIFF
--- a/app/services/submission/plant_trial_finalizer.rb
+++ b/app/services/submission/plant_trial_finalizer.rb
@@ -113,7 +113,7 @@ class Submission::PlantTrialFinalizer
                           plant_trial_id: @plant_trial.id)
       )
 
-      scores_data = (scores || {}).
+      scores_sanitized_data = (scores || {}).
         select{ |_, value| value.present? }.
         map do |col_index, value|
           trait_descriptor = get_nth_trait_descriptor(trait_mapping[col_index])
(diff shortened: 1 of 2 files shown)


app/services/submission/plant_trial_finalizer.rb
```diff
@@ -113,29 +113,49 @@ def create_scoring
...
+
+          trait_score_attributes.merge!(
+            trait_descriptor_id: trait_descriptor.id,
+            score_value: value,
+            plant_scoring_unit_id: new_plant_scoring_unit.id
+          )
+
+          # Ensure consistent order of raw values (by column name)
+          trait_score_attributes.sort.map do |_,v|
+            TraitScore.send(:sanitize, v)
```
kammerer: `send` does not seem to be needed.

kammerer:
:+1:

PR #655, opened 2016-09-01
Assumes query param may not be a hash

Nuanda:
Fixes today's (1.09.16) production issue. Caused by a user playing with possible query settings.


kammerer:
:+1:

Issue #659, opened 2016-09-05
Deal with Zenodo sandbox problem when it's available again


Nuanda:
Apparently, due to some changes in Zenodo Sandbox, our testing API token does not work anymore. Also my account vanished. generating another token on a recreated account results in a 500 error. Posted an email to support and awaiting their response.

Nuanda:
I contacted Zenodo and they fixed the problem. No API changes, just a problem in  their infra.

PR #662, opened 2016-09-05
Updates some traits, creates others and provides trait descriptors to use them in submissions

### LINKED ISSUE #661: Load/update new set of Trait data sent by Annemarie
Some of the traits are already present yet they need update. Accompanying Trait Descriptors should also be created.

Nuanda:
Fixes #661 

Ignoring created/updated Traits in rollback is intentional.

### DIFF
--- a/app/assets/javascripts/data_grid_specific_columns.js.coffee
+++ b/app/assets/javascripts/data_grid_specific_columns.js.coffee
@@ -226,6 +226,13 @@ window.configs =
           else
             data
       ,
+        targets: 'trait_descriptors_scoring_method_column'
+        render: (data, type, full, meta) ->
+          if data
+            data.replace(/(?:\r\n|\r|\n)/g, '<br />');
+          else
+            data
+      ,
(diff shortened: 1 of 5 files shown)


Nuanda:
@kammerer Please take a look. This is needed for this Wed so it would be great if you could do it soon. I just mean - check if i haven't done something totally wrong.

Nuanda:
Broken specs ar related to #659

kammerer:
The only, minor, issue I see is formatting of scoring method, which is shown in the autocomplete field (the asterisks make it look a bit strange).

kammerer:
:+1:

Nuanda:
Yes, this is what I got. Rewriting it is a html list and interpreting user-provided content as html is rather a no go. Only "safe" things I could think of was to use js to replace client-side newlines with `br`s. Committed. Should I keep that change or leave things as they were?

kammerer:
Keep it, IMHO looks a bit better.

PR #663, opened 2016-09-05
PP or PT destroy removes the related submission as well

### LINKED ISSUE #636: Dealing with finalized submissions without the submitted object
It is possible to submit a submissions and then use API to delete the submitted object. BIP should be able to handle that well.

One clean solution would be to destroy related submission (and cascade-destroy its uploads) when a plant trial or a plant population is destroyed.

@kammerer What's your opinion?

Nuanda:
Fixes #636


kammerer:
:+1:

PR #670, opened 2016-09-28
Add manuals

### LINKED ISSUE #668: adding manual files
I am trying to add manual files to the submissions/new  in case of the Plant population and the Plant scoring trial population. 

I did the following thing:
1) created a branch add_manuals
2)added files to new folder public/manuals/upload_population_submission 
                                  and public/manuals/upload_Trial_submission

3)added
= succeed'.' do
            = link_to 'here', 'manuals/upload_population_submission.pdf'.
to the file [submissions.en.yml](https://github.com/TGAC/brassica/blob/3bcfdebc04d275fe4e37dc51ed35604e8bbecfb7/config/locales/submissions.en.yml).
    following this style based on [about.html.haml](https://github.com/TGAC/brassica/blob/eebfffbb631170baebdda6f400d71fe41b09d057/app/views/application/about.html.haml), where you use in line 57
  = succeed '.' do
      = link_to 'in this diagram', 'images/models_complete.png'

4) loaded on local host and only got _= succeed'.' do = link_to 'here', 'manuals/upload_population_submission.pdf'. _ as text instead of a link to the file.

I understand that at the moment in the file it is just a multi-line string. I understand that the file types are different, so maybe it has something to do with this. But I don't know enough to google with the correct search words. So how do would I go about to do this simple edit?

Nuanda:
Fixes #668 

Continuing @teatree1212's work on manuals for manual submission.


Nuanda:
@teatree1212 see my latest commit. There are several things whic construct a _view_, two of them are the _view template_ and _locale_ files.

For this particular view, the main responsible template is `app/views/submissions/new.html.haml`. It is written in the HAML markup and, more or less, decides what things are placed in the view, and what is the layout. It uses a "locale" subsystem, which puts most of the text that appears in the view, to separate files written in YAML markup (there are some advantages of doing so). The file that interests us here is `config/locales/submissions.en.yml`.

What you tried to do was to use HAML markup inside YAML files, which is not possible. YAML is much simpler, it is a kind of "dictionary" file, with (hierarchical) keys to the left, and text (which is supposed to replace the keys in the HAML view) to the right.

However, there are still some tricks we can do with YAML. One is direct HTML (you will see such HTML `<a>` element inserted by me in the "submission.description.common" key). In order to have this HTML rendered correctly, I also needed to add the `.html_safe` call in the respective line of the HAML view. Long story short, it turns off a security measure which deliberately "ruins" all HTML code (in order to prevent evil users ;) injecting malicious HTML code inside BIP). It is safe to turn it off in this specific case as we (i.e. the developers) control what is inside this text.

Another trick is an interpolation: in order to re-use the same YAML text for both "trial" and "population" submissions, I injected the submission type inside the YAML text (using the `, type: s_type` piece of code in the HAML part). Inside YAML, it is used with the `%{type}` markup.

I decided to move the manual link nearer to the end of this piece of information. Do you agree it is better that way or would you still prefer to have it right as the first sentence?

teatree1212:
that looks like a reasonable place to put it.
 Thanks for the explanation!

Nuanda:
@teatree1212 Do you plan to add something here or should we merge it?

teatree1212:
just did some optical improvements.
Ready to merge.

wjurkowski:
I'm wondering if adding more generic information on the top of submission panel could be useful. For instance could we do following:

Submission tutorials: (list of files)
Example of ruby scripts: concrete file (link to a release at github)

PR #671, opened 2016-09-29
Add meta information to the page header

kammerer:
Meta information added as proposed in email conversation.


Nuanda:
Do we need to somehow "register" BIP website in Twitter or Facebook? Should we ask EI social media people to do that for us?

Nuanda:
Also, Wiktor has proposed an additional set of "article:tag" keywords. I guess it does not hurt to add them as well, right?

Nuanda:
A final request would be to change the linked twitter account to https://twitter.com/BrassicaP

kammerer:
I was going to write about this in the email thread. These "article:tag" entries are not valid for BIP, because it is not an article. Besides I understand these are mostly important for Facebook. See http://ogp.me/ for details.

I added all these phrases to "keywords" meta tag, but this is not a solution really, since its value for search engines is very low. The real solution would be to use them in the page text.

kammerer:
We don't need to register the site in anyway on Twitter/Facebook. These tags mainly allow social sites to better present link to BIP if somebody shares it.

kammerer:
Change twitter card to point to "@BrassicaP".

PR #675, opened 2016-10-28
add impact paragraph and external links

teatree1212:
please check


Nuanda:
What do you think about moving the terms of use sentence to the very end?

teatree1212:
thinking about it i Thought it might more sense as standalone sentence after the first paragraph? 
But you can move it to somewhere else if you want. I agree that the sentence where it was was floating around..

teatree1212:
@Nuanda  could you check this before tomorrow please?

PR #686, opened 2016-12-17
DOI links in public space for deposited submissions

### LINKED ISSUE #635: Show assigned DOI inside BIP public space
Currently a BIP user is able to see the Zenodo-assigned DOI when browsing her submissions (which is private space).

It would be good to show this DOI also for other users. In two places:
- on the front-page submission list (if a submission has a DOI)
- inside the (i) Metadata button for PlantPopulations and PlantTrials, if they are submitted via a manual submission, and if that submission has a DOI assigned.

Nuanda:
Fixes #635

### DIFF
--- a/app/views/submissions/show/_status.html.haml
+++ b/app/views/submissions/show/_status.html.haml
@@ -23,7 +23,9 @@
       - else
         Submitted on #{l submission.updated_at, format: :short}
 
-        = submission.decorate.doi
+        - if submission.published?
+          %br
+          DOI: #{submission.decorate.doi}
 
   - else
     %span.label.label-default Draft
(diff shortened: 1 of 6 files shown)


Nuanda:
Thanks. If you try to deposit something in Zenodo sandbox, it returns 500 _most of the time_ (but not _always_). Also when you do that manually via their web tool: the file lands there but when you try to open it, it only shows an error.

When I find some time I'll try to contact them about it. In the meantime I plan to test the functionality manually when deployed to production, with the "publication" Zenodo API call commented out, to see if the production Zenodo environment works as intended.

app/views/submissions/show/_status.html.haml
```diff
@@ -23,7 +23,9 @@
       - else
         Submitted on #{l submission.updated_at, format: :short}
 
-        = submission.decorate.doi
+        - if submission.published?
```
Nuanda: Theoretically this is always true if DOI is not nil, but perhaps you are right - it is good to double check things.

PR #692, opened 2017-01-02
Plant trial export download - enhancement

Nuanda:
- fixed one bug
 - added caching
 - better optimization of trial scoring table building

### DIFF
--- a/app/services/submission/plant_trial_zip_exporter.rb
+++ b/app/services/submission/plant_trial_zip_exporter.rb
@@ -1,12 +1,16 @@
 require 'zip'
 
 class Submission::PlantTrialZipExporter
-  def call(plant_trial)
-    exporter = Submission::PlantTrialExporter.new(
-      OpenStruct.new(submitted_object: plant_trial, user: plant_trial.user)
-    )
+  def call(plant_trial, cache_key)
     compressed_filestream = Zip::OutputStream.write_buffer do |zos|
-      exporter.documents.each do |document_name, content|
+      documents = Rails.cache.fetch(cache_key, expires_in: 300.days) do
+        Rails.logger.info 'MISS MISS MISS'
+        exporter = Submission::PlantTrialExporter.new(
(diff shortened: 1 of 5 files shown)


spec/services/submission/plant_trial_zip_exporter_spec.rb
```diff
@@ -8,7 +8,7 @@
 
   describe "#call" do
     it 'compress three files in the zip file' do
-      csv_data = Submission::PlantTrialZipExporter.new.call(plant_trial).read
+      csv_data = Submission::PlantTrialZipExporter.new.call(plant_trial, 'no_cache').read
```
kammerer: Passing `'no_cache'` seems a bit misleading, since it does not disable cache. It is reused unless cleared between subsequent test runs by some other means. We should probably disable cache globally in test environment (e.g. by switching to `NullStore`) and enable it only for these cases which actually test caching.

app/services/submission/plant_trial_zip_exporter.rb
```diff
@@ -1,12 +1,16 @@
...
 
 class Submission::PlantTrialZipExporter
-  def call(plant_trial)
-    exporter = Submission::PlantTrialExporter.new(
-      OpenStruct.new(submitted_object: plant_trial, user: plant_trial.user)
-    )
+  def call(plant_trial, cache_key)
     compressed_filestream = Zip::OutputStream.write_buffer do |zos|
-      exporter.documents.each do |document_name, content|
+      documents = Rails.cache.fetch(cache_key, expires_in: 300.days) do
```
kammerer: I don't like mixing caching logic and building zip file, but I am not sure how to split it.

Nuanda: I agree, though caching a stream is not desirable. Hence I opted for merging these two aspects in this single service.

Nuanda:
One thing done, the other left as it is. Merging.

kammerer:
:+1:

PR #694, opened 2017-01-05
Plant Population submission wizard overhaul

### LINKED ISSUE #693: PP submission: move population_type to step1 and parent lines to step2
In line with newest propositions. Related to #494

Nuanda:
Fixes #693 

Partially addresses #494 

Initial commit, WIP.

@kammerer Tomasz, I need your assistance with this, to have it completed rather soon. The step03 of the population submission features now a file upload widget. You can upload a CSV file there and a parser gets plant lines information from the file and puts it into the plant_line_list property of the submission content.

The current problem is that the form view is not updated. I would like to keep the current manual plant line definition, so the file upload should simply complement it with another method of plant line information input. So, after the upload and parsing, the plant line list input should be populated with data from the CSV file (appended to whatever the user put there earlier).

Currently, after file upload and parsing, I need to reload the view to see the new lines in the form - the idea is to have them there right after the file was parsed. Could you help with this?


Nuanda:
@kammerer Tomasz, one more comment. If you look at PlantLineParser you'll see disabled lines 63-64. When you modify the process so the file upload appends (instead of replacing) new PLs to the list, I think you should ignore (and inform the user about that in the parsing log) all PLs with repeated PL names. In the (probably frequent) case when the user re-uploads improved CSV, we may simply require her to clear the PL list input first. But this way we won't run into troubles during the finalization procedure.

kammerer:
@Nuanda I updated javascript upload handler to append new PLs to the list as you requested. I have not modified the parser yet.

Nuanda:
Somehow it does not work me - a parsed line does not appear in the plant lines list input (again, I need to reload the page to see that). Don't have much time for testing now, will do more tests in the evening.

kammerer:
Yeah, apparently I didn't test the last change... Anyway, I am working on the missing parts now, so I will fix this too.

Nuanda:
Yes, it looks better now.

Nuanda:
I will add missing specs for the PPSubmFinalizer.

app/forms/submissions/population/step03_content_form.rb
```diff
@@ -103,6 +92,11 @@ def new_plant_lines
           plant_line_list.include?(plant_line.plant_line_name)
         end
       end
+
+      def upload
+        upload = Submission::Upload.plant_lines.find_by(id: upload_id)
+        SubmissionUploadDecorator.decorate(upload) if upload
```
Nuanda: @kammerer You have added a specialised decorator. Should this line be updated?

kammerer: It should be ok to update it.

kammerer:
@Nuanda Is there anything else I should do here right now?

app/forms/submissions/population/step03_content_form.rb
Nuanda: Done.

Nuanda:
@kammerer Yes, there is. Please review the entire thing.

spec/services/submission/plant_population_finalizer_spec.rb
```diff
@@ -12,26 +12,46 @@
 
   context 'given submission with valid content' do
     let(:new_plant_lines_attrs) {
-      attributes_for_list(:plant_line, 2).map { |attrs|
+      attributes_for_list(:plant_line, 2).map.with_index { |attrs,i|
```
kammerer: I don't like to pick on syntactic details but please, add space after comma.

PR #696, opened 2017-01-11
Extend accession uniqueness

### LINKED ISSUE #641: Extend PT scoring template with PA.year_produced and make it obligatory
@teatree1212: The year_produced information should be added to the accession in the plant_trial_scoring submission template, as this can explain variation in the data.

### LINKED ISSUE #656: Extend the PA uniqueness requirement
For PAs two following combination of _three_ fields should be required to be unique: `plant_accession`, `originating_organisation` and `year_produced`. All other uniqueness rules and DB indices should be removed for the PA model and this one should be added instead.

Implement only after #641.

nowakowski:
Fixes #656 
Fixes #641


Nuanda:
@kammerer Tomasz, please have a look as well. IMHO it is mergeable, apart from the single remark I made in my review.

PR #702, opened 2017-02-10
Add a script for fetching trial scoring in TSV format for YorkAT

kammerer:
A version of Annemarie's script for fetching trial scoring data and formatting it in such a way that it can be used directly in collaborator's association pipeline.

### DIFF
--- a/public/client_example/trial_analysis_YorkAT.rb
+++ b/public/client_example/trial_analysis_YorkAT.rb
@@ -1,0 +1,193 @@
+require 'csv'
+require 'json'
+require 'uri'
+require 'net/http'
+require 'net/https'
+
+# A BIP Ruby client which, given a Plant Trial name, fetches:
+#
+#  * all Trait Scores (trait measurement values)
+#  * all corresponding Trait Descriptors
+#  * Line Names
+#
+# and produces a TSV to standard output, with each Plant Scoring Unit represented by a row.
+
+if ARGV.size < 2
+  STDERR.puts 'Not enough arguments. Usage:'
+  STDERR.puts ' > ruby trial_analysis.rb plant_trial_name api_key'
+  exit 1
+end
+
+@client = Net::HTTP.new('bip.earlham.ac.uk', 443)
+@client.use_ssl = true
+@headers = {
+  'Content-Type' => 'application/json',
+  'X-BIP-Api-Key' => ARGV[1]
+}
+
+def pluralize_class(class_name)
+  class_name == 'plant_variety' ? 'plant_varieties' : "#{class_name}s"
+end
+
+def call_bip(request)
+  response = @client.request(request)
+  JSON.parse(response.body).tap{ |parsed| test_error(parsed) }
+end
+
+def test_error(response)
+  unless response['reason'].nil?
+    STDERR.puts "Encountered the following BIP request error."
+    STDERR.puts response['reason']
+    exit 1
+  end
+end
+
+outputs = {}
+
+STDERR.puts '1. Finding the Plant Trial'
+
+trial_encoded_name = URI.escape ARGV[0]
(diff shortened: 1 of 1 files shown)


public/client_example/trial_analysis_YorkAT.rb
```diff
@@ -0,0 +1,193 @@
...
+
+
+STDERR.puts '8. Generating output TSV to STDOUT'
+
+output_string = CSV.generate(col_sep: "\t") do |csv|
+  csv << ["<Trait>"] + trait_descriptors.map{ |td| td['trait']['name'].gsub(/\s+/, '.') }
+  outputs.
+    select do |_, data|
+      trait_descriptors.select { |td|
+        data['trait_scores'].detect { |ts| ts['trait_descriptor_id'] == td['id'] }
```
Nuanda: I assume this is here to exclude PSUs which do not provide full scoring info. Please note this produces an empty output file to trials like this one: https://bip.earlham.ac.uk/trial_scorings/2. But i understand this is intentional.

public/client_example/trial_analysis_YorkAT.rb
```diff
@@ -0,0 +1,193 @@
...
+      }.
+      count == trait_descriptors.count
```
Nuanda: Please, please change this code ;)! I was thinking for an entire minute where the variable `count` is being defined... Then, I saw the dot. I think having this one local variable assigned in the previous line, and compared here, would yield more readable code.

public/client_example/trial_analysis_YorkAT.rb
```diff
@@ -0,0 +1,193 @@
...
+    outputs[plant_scoring_unit['scoring_unit_name']]['plant_accession'] = plant_accessions.detect { |pa| pa['id'] == plant_scoring_unit['plant_accession_id']}
+  end
+  page += 1
+end
+
+
+STDERR.puts '8. Generating output TSV to STDOUT'
+
+output_string = CSV.generate(col_sep: "\t") do |csv|
+  csv << ["<Trait>"] + trait_descriptors.map{ |td| td['trait']['name'].gsub(/\s+/, '.') }
```
Nuanda: Why use this specific "<Trait>" column header here? This is the plant line name column, so this kind of name would be preferable, so the users who find the generated tsv file after half a year on their disks do not wonder, what values are actually presented in the 1st column.

If this is not an option (e.g. some external tool requires this particular header), maybe consider a line of comment above the header?

Similar question regarding the strange "." notation for trait names. Is that required by some external tool? If not, perhaps quoting the names would be sufficient?

public/client_example/trial_analysis_YorkAT.rb
```diff
@@ -0,0 +1,193 @@
...
+          }
+
+          ts['score_value'] if ts
+        }
+    end
+end
+
+puts output_string
+
+STDERR.puts '6. Finished'
```
Nuanda: It's not "6." anymore :).

kammerer:
I am not sure about #681 but I think Annemarie does not work on it any longer anyway. I would not however delete that branch as of yet because the original version of the script may be useful (although both of them could use some tidying).

public/client_example/trial_analysis_YorkAT.rb
kammerer: This is based on an example file I was given by Wiktor. I do not have any more information, so I just tried to make the output as similar as possible. I will add some comments regarding this formatting.

public/client_example/trial_analysis_YorkAT.rb
kammerer: Yes, this is intentional, however needs to be confirmed with the collaborator.

public/client_example/trial_analysis_YorkAT.rb
kammerer: OK

Nuanda:
:+1:

PR #709, opened 2017-03-31
Fist implementation of germplasm-search BrAPIv1 call

CarlosHorro:
As this is my first commit here, I would specially like to know if I'm structuring things well, I would have to change the way I'm doing them, or I'm missing (or not reusing) any required functionality of the application, in order to continue working in this way or not.

Finally I've created pure SQL as there are (and there will be) some specific aspects of the returned data (many entities involved but not so many fields, some entities referenced by different relations, etc.) which need some manual treatment and it can be done with a lot better performance directly in this way.
Relations between BrAPI and BIP fields are based (as far as it has been possible yet) on Annemarie's mapping work:
https://docs.google.com/spreadsheets/d/1UicLL_FghRc8X-rVpAwlDfxQdJoRaoI4RR6icpN7J-E/edit#gid=0

To be mentioned that, except a BrAPI mandatory return field as 'instituteCode' (to be studied next), and one parameter as 'germplasmPUI', germplasm-search call is currently working.

### DIFF
--- a/app/controllers/brapi/v1/germplasm_controller.rb
+++ b/app/controllers/brapi/v1/germplasm_controller.rb
@@ -1,0 +1,250 @@
+class Brapi::V1::GermplasmController < Brapi::BaseController
+
+  attr_accessor :request_params, :user
+
+  before_filter :authenticate_api_key!
+
+
+  rescue_from 'ActiveRecord::InvalidForeignKey' do |exception|
+    message = exception.message.split("\n").try(:second)
+    attribute = message ? message[14..-1].split(')')[0] : ''
+    render json: { errors: { attribute: attribute, message: message } }, status: 422
+  end
+
+  rescue_from ActionController::ParameterMissing do |exception|
+    render json: { errors: { attribute: exception.param, message: exception.message } }, status: 422
+  end
+
+
+
+
+ def search
+    # accepted params: germplasmPUI , germplasmDbId , germplasmName , pageSize , page
+    # domainParameters = ['germplasmPUI','germplasmDbId','germplasmName']
+    # totalParameters = domainParameters + ['pageSize','page']
+    
+    if !params['germplasmPUI'].present? && !params['germplasmDbId'].present? && !params['germplasmName'].present?
+      render json: { reason: 'Resource not found' }, status: :not_found
+    else 
+      result_object = germplasm_query(params, false)
+      records = result_object.values
+      
+      
+      if records.nil?
+        render json: { reason: 'Resource not found' }, status: :not_found
+      else
+        if records.size == 1 && records[0]!=nil && records[0] != api_key.user_id && records[1] 
+          render json: { reason: 'This is a private resource of another user' }, status: :unauthorized
+        else
+          json_result_array = []
+          
+          # any programmatic data manipulation can be done here
+          result_object.each do |row|
+            row = JSON.parse(row["row_to_json"])
+            puts row
+            if (!row[:user_id] || (row[:user_id]!=nil && row[:user_id] != api_key.user_id && row[:published] ))
+              json_result_array << row
+            end
+          end
+          
+          # pagination data returned
+          page = get_page
+          pageSize = get_pageSize 
+          
+          result_count_object = germplasm_query(params, true)
+          totalCount = result_count_object.values.first[0].to_i
+          totalPages = (totalCount/pageSize.to_f).ceil
+          
+          json_response = { 
+            'metadata' => json_metadata(pageSize, page, totalCount, totalPages),
+            'result' => json_result_array
+          }
+         
+          render json: json_response, :except => ["id", "user_id", "created_at", "updated_at","total_entries_count"]
+        end
+      end
+      
+    
+    end
+  end
+
+
+
+
+  private
+
+
+
+  # processing result: https://github.com/plantbreeding/API/blob/master/Specification/Germplasm/GermplasmSearchGET.md 
+  # All possible fields to return:
+  # germplasmDbId, defaultDisplayName, accessionNumber, germplasmName, germplasmPUI, pedigree, 
+  # seedSource, synonyms, commonCropName, instituteCode, instituteName, biologicalStatusOfAccessionCode, 
+  # countryOfOriginCode, typeOfGermplasmStorageCode, genus, species, speciesAuthority, subtaxa, subtaxaAuthority, 
+  # donors, acquisitionDate
+
+  def germplasm_query(params, count_mode)
+    # where conditions
+    where_query = " "
+    where_atts = []
+    where_atts_count = 0
+    
+    # TO BE DEFINED
+    #if params['germplasmPUI'].present? 
+    
+    if params['germplasmDbId'].present?   # plant_accessions.plant_accession
+      where_atts_count+= 1
+      where_query = where_query + "where plant_accessions.plant_accession = $"+where_atts_count.to_s
+      where_atts<< params['germplasmDbId']
+    end
+     
+    if params['germplasmName'].present?   # plant_lines.plant_variety_name or plant_common_name
+      where_atts_count+= 1
+      where_query = where_query + where_atts_count==1?" where":" and" + " (plant_lines.plant_variety_name = $"+where_atts_count.to_s+" OR 
+       plant_lines.plant_common_name = $"+where_atts_count.to_s+")"
+      where_atts<< params['germplasmName']
+    end
+    
+    
+    select_query = ""
+    
+    select_query = "SELECT DISTINCT ON (plant_accessions.id)
+    plant_accessions.id,
+    plant_accessions.user_id,
+    plant_accessions.published,
+    plant_accessions.plant_accession as \"germplasmDbId\", 
+    case 
+      when plant_lines.common_name != null then plant_lines.common_name
+      else plant_lines.plant_variety_name
+    end as \"defaultDisplayName\",
+    plant_accessions.plant_accession as \"accessionNumber\",
+    plant_lines.common_name as \"commonCropName\",
+    null as \"instituteCode\",
+    plant_accessions.originating_organisation as \"instituteName\",
+    plant_populations.population_type as \"biologicalStatusOfAccessionCode\",
+    case 
+      when countries_registered_from_lines.id != null then countries_registered_from_lines.country_code
+      else countries_registered_from_accessions.country_code
+    end as \"countryOfOriginCode\",
+    'Brassica' as \"genus\",
+    regexp_replace(taxonomy_terms.name, '( var.).*$', '') as \"species\",  
+    taxonomy_terms.name as \"subtaxa\" 
+    "
+    # TODO : instituteCode mandatory : to be implemented. To review instituteName
+    
+    
+    joins_query = "
+    FROM plant_accessions
+    INNER JOIN plant_lines ON plant_accessions.plant_line_id = plant_lines.id
+    INNER JOIN plant_population_lists ON plant_lines.id = plant_population_lists.plant_line_id
+    INNER JOIN plant_populations ON plant_population_lists.plant_population_id = plant_populations.id
+    LEFT JOIN plant_varieties AS plant_varieties_from_lines ON plant_lines.plant_variety_id = plant_varieties_from_lines.id 
+    LEFT JOIN plant_varieties AS plant_varieties_from_accessions ON plant_accessions.plant_variety_id = plant_varieties_from_accessions.id 
+    LEFT JOIN plant_variety_country_registered AS plant_variety_country_registered_from_lines ON plant_varieties_from_lines.id = plant_variety_country_registered_from_lines.plant_variety_id 
+    LEFT JOIN plant_variety_country_registered AS plant_variety_country_registered_from_accessions ON plant_varieties_from_accessions.id = plant_variety_country_registered_from_accessions.plant_variety_id 
+    LEFT JOIN countries AS countries_registered_from_lines ON plant_variety_country_registered_from_lines.country_id = countries_registered_from_lines.id 
+    LEFT JOIN countries AS countries_registered_from_accessions ON plant_variety_country_registered_from_accessions.country_id = countries_registered_from_accessions.id 
+    LEFT JOIN taxonomy_terms ON plant_lines.taxonomy_term_id = taxonomy_terms.id
+    "
+    
+    
+    # order
+    order_query =  " ORDER BY plant_accessions.id desc"
+    
+    if count_mode
+      total_query = "SELECT COUNT(*) FROM ("+select_query + joins_query + where_query + order_query+") AS total_entries_count"
+
+      puts total_query
+      result_object = execute_statement(total_query, where_atts)
+    else
+      # pagination
+      page = get_page
+      pageSize = get_pageSize
+      pagination_query = pagination_query(page, pageSize)
+      
+      total_query = select_query + joins_query + where_query + order_query + pagination_query
+      
+      json_wrapping_query = "SELECT row_to_json(row) from ("+total_query+") row"
+      
+      result_object = execute_statement(json_wrapping_query, where_atts)
+    end
+    
+    records = result_object.values
+    result_object
+  end
+
+
+  def pagination_query(page, pageSize)
+    return " LIMIT "+pageSize.to_s+" OFFSET "+((page-1)*pageSize).to_s  
+  end
+  
+  
+  def get_page
+    return (params[:page].to_i != 0)?params[:page].to_i : 1
+  end
+  
+  def get_pageSize
+    return (params[:pageSize].to_i != 0)?params[:pageSize].to_i : 1000 
+  end
+  
+  def json_metadata(pageSize, currentPage, totalCount, totalPages)
+    json_metadata = {
+      status: [],
+      files: [],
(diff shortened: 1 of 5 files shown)


kammerer:
@CarlosHorro Please, add some [request specs](https://www.relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec) to the PR. There are things to change in the code but that should not affect the way new API is visible from the outside too much.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+  end
+
+  rescue_from ActionController::ParameterMissing do |exception|
+    render json: { errors: { attribute: exception.param, message: exception.message } }, status: 422
+  end
+
+
+
+
+ def search
```
Nuanda: One space too few?

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+      pagination_query = pagination_query(page, pageSize)
+      
+      total_query = select_query + joins_query + where_query + order_query + pagination_query
+      
+      json_wrapping_query = "SELECT row_to_json(row) from ("+total_query+") row"
+      
+      result_object = execute_statement(json_wrapping_query, where_atts)
+    end
+    
+    records = result_object.values
```
Nuanda: Is `records` used anywhere?

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+      else
+        if records.size == 1 && records[0]!=nil && records[0] != api_key.user_id && records[1] 
+          render json: { reason: 'This is a private resource of another user' }, status: :unauthorized
+        else
+          json_result_array = []
+          
+          # any programmatic data manipulation can be done here
+          result_object.each do |row|
+            row = JSON.parse(row["row_to_json"])
+            puts row
```
Nuanda: Please remove debug outputs when making a PR to master.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+            if (!row[:user_id] || (row[:user_id]!=nil && row[:user_id] != api_key.user_id && row[:published] ))
```
Nuanda: I am not sure but will this accept unpublished records owned by the user who has issued the query?

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+      render json: { reason: 'Resource not found' }, status: :not_found
+    else 
+      result_object = germplasm_query(params, false)
+      records = result_object.values
+      
+      
+      if records.nil?
+        render json: { reason: 'Resource not found' }, status: :not_found
+      else
+        if records.size == 1 && records[0]!=nil && records[0] != api_key.user_id && records[1]
```
Nuanda: Please consistently apply spaces before and after operators (!=).

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+            row = JSON.parse(row["row_to_json"])
+            puts row
+            if (!row[:user_id] || (row[:user_id]!=nil && row[:user_id] != api_key.user_id && row[:published] ))
+              json_result_array << row
+            end
+          end
+          
+          # pagination data returned
+          page = get_page
+          pageSize = get_pageSize
```
Nuanda: Please use underscored_names for Ruby variables and camelCaseNames for JSON document only.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+          result_count_object = germplasm_query(params, true)
+          totalCount = result_count_object.values.first[0].to_i
+          totalPages = (totalCount/pageSize.to_f).ceil
+          
+          json_response = { 
+            'metadata' => json_metadata(pageSize, page, totalCount, totalPages),
+            'result' => json_result_array
+          }
+         
+          render json: json_response, :except => ["id", "user_id", "created_at", "updated_at","total_entries_count"]
```
Nuanda: Consistently apply spaces between array elements.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+    
+    if params['germplasmDbId'].present?   # plant_accessions.plant_accession
+      where_atts_count+= 1
+      where_query = where_query + "where plant_accessions.plant_accession = $"+where_atts_count.to_s
+      where_atts<< params['germplasmDbId']
+    end
+     
+    if params['germplasmName'].present?   # plant_lines.plant_variety_name or plant_common_name
+      where_atts_count+= 1
+      where_query = where_query + where_atts_count==1?" where":" and" + " (plant_lines.plant_variety_name = $"+where_atts_count.to_s+" OR
```
Nuanda: Please split that line so it is easier to read.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+    end
+    
+    
+    select_query = ""
+    
+    select_query = "SELECT DISTINCT ON (plant_accessions.id)
+    plant_accessions.id,
+    plant_accessions.user_id,
+    plant_accessions.published,
+    plant_accessions.plant_accession as \"germplasmDbId\",
```
Nuanda: I think you are able to use`'` to delimit Ruby strings, so you don't have to escape `"` inside.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+    LEFT JOIN countries AS countries_registered_from_accessions ON plant_variety_country_registered_from_accessions.country_id = countries_registered_from_accessions.id 
+    LEFT JOIN taxonomy_terms ON plant_lines.taxonomy_term_id = taxonomy_terms.id
+    "
+    
+    
+    # order
+    order_query =  " ORDER BY plant_accessions.id desc"
+    
+    if count_mode
+      total_query = "SELECT COUNT(*) FROM ("+select_query + joins_query + where_query + order_query+") AS total_entries_count"
```
Nuanda: I am not sure you need to `order` things when you just need the `count`.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+    where_atts = []
+    where_atts_count = 0
+    
+    # TO BE DEFINED
+    #if params['germplasmPUI'].present? 
+    
+    if params['germplasmDbId'].present?   # plant_accessions.plant_accession
+      where_atts_count+= 1
+      where_query = where_query + "where plant_accessions.plant_accession = $"+where_atts_count.to_s
+      where_atts<< params['germplasmDbId']
```
Nuanda: If I am correct, you are putting user-provided parameters "as are" inside the query string. If this is the case, please add some code to check the `params` content against code injection.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+
+
+  
+  def execute_statement(sql, atts)
+      connection = ActiveRecord::Base.connection.raw_connection
+
+      connection.prepare('brapi_statement', sql)
+      
+      if( (atts != nil) && !(atts.empty?) )
+        results = connection.exec_prepared("brapi_statement", atts)
```
Nuanda: OK, I see you pass the user-provided attribute values to `exec_prepared` so I think this is safe, code injection wise.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -0,0 +1,250 @@
...
+      else
+        results = connection.exec_prepared("brapi_statement")
+      end
+      connection.exec("DEALLOCATE brapi_statement")
+
+      if results.present?
+          return results
+      else
+          return nil
```
kammerer: Returning `nil` here seems to be an error as you later call `.values` on returned object. It probably makes sense to return empty result object.

kammerer:
I would argue in favor of moving SQL-related code to a separate class.

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: I like to visually separate different areas, like 'search' from
rescue_from ActionController::ParameterMissing do |exception|
...

Nuanda: No - I mean there is (I think) only one space indentation before the "def" keyword :).

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: As BrAPI doesn't currently support ORCID, these lines are being modifying. 
New version will be something like:
if (!row[:user_id] || row[:published] )
as we only should return published datasets, or datasets without an author... if I'm not wrong...

Nuanda: Sounds correct.

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: Ok, I will review the entire file...

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: It isn't, I will remove it.

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: Is not necessary to do extra checks to 'params', then?

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: Totally right, changing it...

app/controllers/brapi/v1/germplasm_controller.rb
Nuanda: I think not, but since the matter is rather important, I'd like to ask @kammerer for a second opinion on this.

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: Um... yes, although I think I will also need to scape ' in other areas:
select_query = 'SELECT DISTINCT ON (plant_accessions.id)
    plant_accessions.id,
    plant_accessions.user_id,
    plant_accessions.published,
    plant_accessions.plant_accession as "germplasmDbId", 
    case 
      when plant_lines.common_name != null then plant_lines.common_name
      else plant_lines.plant_variety_name
    end as "defaultDisplayName",
    plant_accessions.plant_accession as "accessionNumber",
    plant_lines.common_name as "commonCropName",
    null as "instituteCode",
    plant_accessions.originating_organisation as "instituteName",
    plant_populations.population_type as "biologicalStatusOfAccessionCode",
    case 
      when countries_registered_from_lines.id != null then countries_registered_from_lines.country_code
      else countries_registered_from_accessions.country_code
    end as "countryOfOriginCode",
    \'Brassica\' as "genus",
    regexp_replace(taxonomy_terms.name, \'( var.).*$\', \'\') as "species",  
    taxonomy_terms.name as "subtaxa" 
    '

CarlosHorro: Um Yes, although I think I will also need to escape ' in other parts (regular expression and Brassica literal):

`select_query = 'SELECT DISTINCT ON (plant_accessions.id)
    plant_accessions.id,
    plant_accessions.user_id,
    plant_accessions.published,
    plant_accessions.plant_accession as "germplasmDbId", 
    case 
      when plant_lines.common_name != null then plant_lines.common_name
      else plant_lines.plant_variety_name
    end as "defaultDisplayName",
    plant_accessions.plant_accession as "accessionNumber",
    plant_lines.common_name as "commonCropName",
    null as "instituteCode",
    plant_accessions.originating_organisation as "instituteName",
    plant_populations.population_type as "biologicalStatusOfAccessionCode",
    case 
      when countries_registered_from_lines.id != null then countries_registered_from_lines.country_code
      else countries_registered_from_accessions.country_code
    end as "countryOfOriginCode",
    \'Brassica\' as "genus",
    regexp_replace(taxonomy_terms.name, \'( var.).*$\', \'\') as "species",  
    taxonomy_terms.name as "subtaxa" 
    '`

CarlosHorro:
I agree with moving SQL-related code to a separate class, but what folder of the app do you think is the best to place that new class? /services?

app/controllers/brapi/v1/germplasm_controller.rb
kammerer: You can also use so called "heredoc" strings and you will not need to escape quotes at all.

kammerer: E.g.
```ruby
select_query = <<-SQL.strip_heredoc
  SELECT stuff
  FROM table
SQL
```

Using `strip_heredoc` method is an option (it comes with RoR, not Ruby itself).

app/controllers/brapi/v1/germplasm_controller.rb
kammerer: I think it is safe but I will have a closer look later.

kammerer:
I think `/services/brapi` is fine.

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: Great thanks! That is very useful :-)

app/controllers/brapi/v1/germplasm_controller.rb
CarlosHorro: Understood :-)

CarlosHorro:
By the way, I don't understand why Travis checks are failing with classes I haven't touched at all...

Nuanda:
@CarlosHorro Carlos, there failures are not due to your changes. They are regarding Zenodo and ORCiD integration. I'll look into that when I have some time, but you may proceed with the review of your code.

CarlosHorro:
Last changes solve all mentioned problems?

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -2,68 +2,64 @@ class Brapi::V1::GermplasmController < Brapi::BaseController
 
   attr_accessor :request_params, :user
 
-  before_filter :authenticate_api_key!
+  # Authentication via ORCID is currently unsupported by BrAPI.
```
kammerer: To clarify, BIP API uses internal authentication. ORCID is used only for the UI.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -2,68 +2,64 @@ class Brapi::V1::GermplasmController < Brapi::BaseController
...
+  # before_filter :authenticate_api_key!
```
kammerer: Please, do not leave commented out code like this.

spec/controllers/brapi/v1/germplasm_controller_spec.rb
```diff
@@ -0,0 +1,89 @@
+require 'rails_helper'
+require 'pp'
+
+RSpec.describe Brapi::V1::GermplasmController do
+  
+  # BrAPI doesn't use ORCID authentication system and WP7 hasn't agreed yet on an alternative.
```
kammerer: Please, correct this comment and move it to the controller source file.

app/controllers/brapi/v1/germplasm_controller.rb
```diff
@@ -2,68 +2,64 @@ class Brapi::V1::GermplasmController < Brapi::BaseController
...
+          params['germplasmPUI'],params['germplasmDbId'], params['germplasmName'], page, page_size, true)
+        total_count = result_count_object.values.first[0].to_i
+        total_pages = (total_count/page_size.to_f).ceil
+        
+        json_response = { 
+          'metadata' => json_metadata(page_size, page, total_count, total_pages),
+          'result' => json_result_array
+        }
+       
+        render json: json_response, :except => ["id", "user_id", "created_at", "updated_at", "total_entries_count"]
```
kammerer: Currently preferred hash syntax is `{ foo: "bar" }` instead of `{ :foo => "bar" }`.

app/services/brapi/v1/germplasm_queries.rb
```diff
@@ -0,0 +1,142 @@
+require 'singleton'
+class Brapi::V1::GermplasmQueries
+  include Singleton
+
+  def initialize
+    # Currently only one connection opened. 
+    # TODO: For the future would be good to create a connection pool.
+    @connection = ActiveRecord::Base.connection.raw_connection
```
kammerer: AFAIK ActiveRecord provides connection pooling internally.

CarlosHorro:
As I think I've taken into consideration all comments to the pull request, I've merged it with the brassica/BRAPI branch.
I have almost finished another 2 calls which I will try to include into a new pull request during next days.

PR #719, opened 2017-05-12
BrAPI v1: studies-search and studies/:id calls, first implementation

CarlosHorro:
Following the same mechanisms than germplasm-search call, I've implemented:
studies-search call (POST)
studies/:id call (GET)

Although there are still some missing fields (they will have to be added with time and, probably, together with MIAPPE new fields), I think it's a good base to work with.

Last recommended call to be implemented, phenotypes-search, will also probably have to wait until MIAPPE adoption.

If possible, It would be great to have these calls reviewed soon in order to show all these 3 first BrAPI v1 calls working by the end of this month on the next BYOD BrAPI meeting in Ghent, please.


app/controllers/brapi/v1/studies_controller.rb
```diff
@@ -0,0 +1,178 @@
...
+        }
+       
+        render json: json_response, except: ["id", "user_id", "created_at", "updated_at", "total_entries_count"]
+      end
+    
+    end
+  end
+
+
+  def shown
```
CarlosHorro: yes, changing...

app/controllers/brapi/v1/studies_controller.rb
```diff
@@ -0,0 +1,178 @@
...
+          # To check authentication and ownership when ORCID is supported by BrAPI
+          # We currently only retrieve public records. This is already done at query level
+          #if (!row[:user_id] || row[:published] )
+          json_result_array << row
+          #end
+        end
+        
+        json_response = { 
+          metadata: json_metadata(0, 0, records.size, 0),
+          result: json_result_array
```
CarlosHorro: Uf... this is complicated as BrAPI is not always very consistent as it input/output structure sometimes depends on the specific call.
As you say, in SeachStudies they show the 'data' area:
https://github.com/plantbreeding/API/blob/master/Specification/Studies/SearchStudies.md

But they don't include it in studies/:id  : 
https://github.com/plantbreeding/API/blob/master/Specification/Studies/StudyDetails.md

So... I will try to contact with them in order to know what is the correct way of doing this...

app/controllers/brapi/v1/studies_controller.rb
```diff
@@ -0,0 +1,178 @@
...
+  def get_page_size
+    return (params[:pageSize].to_i != 0)?params[:pageSize].to_i : 1000 
+  end
+  
+  def json_metadata(page_size, current_page, total_count, total_pages)
+    json_metadata = {
+      status: [],
+      files: [],
+      pagination: {
+        pageSize: page_size, # old 'per_page'
```
CarlosHorro: I wrote that as a reminder of its "equivalence" with the "per_page" property in CollectionDecorator. For me was useful although maybe improvable...

kammerer: I think it is rather obvious what `pageSize`, but in this context it is not clear what is "old 'per_page'". So, for me at least, it is confusing :)

app/services/brapi/v1/studies_queries.rb
```diff
@@ -0,0 +1,289 @@
...
+      order_query += " countries.country_name "
+    when "programName"
+      order_query += " plant_trials.project_descriptor "
+    else
+      order_query += " plant_trials.id "
+    end
+    return order_query    
+  end
+
+  def get_posgresql_where_condition(field, value, condition_number )
```
CarlosHorro: I included the mention to postgresql because I used a syntax which could be highly dependent on it (there, and in other methods)... so, do I remove that mention in all of them?

app/services/brapi/v1/studies_queries.rb
```diff
@@ -0,0 +1,289 @@
...
+    
+      plant_trials.user_id,
+      plant_trials.id as "studyDbId", 
+      plant_trials.plant_trial_name as "name", 
+      plant_trials.id as "trialDbId", 
+      plant_trials.plant_trial_name as "trialName",   
+      plant_trials.trial_year as "seasons",           
+      countries.id as "locationDbId",
+      countries.country_name as "locationName",
+      plant_scoring_units.plant_accession_id,
```
CarlosHorro: Totally right, I forgot them there after some tests.

app/services/brapi/v1/studies_queries.rb
kammerer: PostgreSQL is an explicit dependency, so I do not see the need to mention it here.  

As a side note, `DISTINCT ON` is specific to PostgreSQL, but there is no mention in the method name ;)

app/controllers/brapi/v1/studies_controller.rb
CarlosHorro: I've checked in their mock server that they use "data" structure when that call can return more than one "data object", and don't use it with calls which only retrieves 0 or 1 (list vs. :id calls). So I will apply the same criteria.

spec/controllers/brapi/v1/studies_controller_spec.rb
```diff
@@ -0,0 +1,167 @@
+require 'rails_helper'
+require 'pp'
+
+RSpec.describe Brapi::V1::StudiesController do
```
CarlosHorro: One question... I'm starting with the request spec and I would like to execute it alone, but an error appears:
rspec spec/requests/brapi/v1/brapi_spec.rb 
/Users/horroc/Documents/Proyects GIT/brassica-own-fork/brassica/spec/rails_helper.rb:63:in `block in <top (required)>': uninitialized constant RSpecHtmlMatchers (NameError) ...

Is it possible to execute only one rspec?

kammerer: Sure, it should be possible. Can you share the code, e.g. push it to a temporary branch in your repo?

app/services/brapi/v1/studies_queries.rb
```diff
@@ -0,0 +1,289 @@
...
+    joins_query = 
+    "FROM plant_trials
+    LEFT JOIN plant_scoring_units ON plant_trials.id = plant_scoring_units.plant_trial_id
+    LEFT JOIN plant_accessions ON plant_scoring_units.plant_accession_id = plant_accessions.id
+    LEFT JOIN countries ON plant_trials.country_id = countries.id
+    "
+    
+    total_query = select_query + joins_query + where_query
+    
+    json_wrapping_query = "SELECT row_to_json(row) from ("+total_query+") row"
```
CarlosHorro: Well, row_to_json returns a json structure but into a String and without a good formatting, like this:
"{\"id\":254,\"user_id\":null,\"germplasmDbId\":\"hzau2003_TN077_a03\",\"defaultDisplayName\":null,\"accessionNumber\":\"hzau2003_TN077_a03\",\"commonCropName\":null,\"instituteCode\":null,\"instituteName\":\"HZAU\",\"biologicalStatusOfAccessionCode\":\"DH segregating\",\"countryOfOriginCode\":null,\"genus\":\"Brassica\",\"species\":\"Brassica napus\",\"subtaxa\":\"Brassica napus\"}"

Using JSON.parse returns a Hash which can be rendered directly by
render json: json_response.to_json
with good formatting and without weird symbols:
{"id"=>254,
 "user_id"=>nil,
 "germplasmDbId"=>"hzau2003_TN077_a03",
 "defaultDisplayName"=>nil,
 "accessionNumber"=>"hzau2003_TN077_a03",
 "commonCropName"=>nil,
 "instituteCode"=>nil,
 "instituteName"=>"HZAU",
 "biologicalStatusOfAccessionCode"=>"DH segregating",
 "countryOfOriginCode"=>nil,
 "genus"=>"Brassica",
 "species"=>"Brassica napus",
 "subtaxa"=>"Brassica napus"}

CarlosHorro: To remark that it seems github has removed the \ " from all keys of the first example.

kammerer: What I meant was why not return a hash directly from the DB and then transform it to json?

CarlosHorro: Now I understand. I didn't do that because I had some problems with the final coding that I didn't have in this way, but I've just realised that I was doing the json transformation in a wrong way.
I will update all this code.

spec/controllers/brapi/v1/studies_controller_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+  
+  context 'BrAPI without requiring authentication' do
+    let(:u1) { create :user }
+    before { sign_in u1 }
+    
+    describe '#search' do
```
CarlosHorro: I've created some of them in last commit based on previous controller specs. Do I remove these controller ones?

spec/controllers/brapi/v1/studies_controller_spec.rb
CarlosHorro: it's testable in today's commit.

spec/controllers/brapi/v1/studies_controller_spec.rb
kammerer: These controller specs are now redundant and should be removed. However, some of the specs for studies API fail, please check what's wrong. Other than that, I think you can merge.

PR #720, opened 2017-05-12
Plant trial based GWAS

kammerer:
This PR makes it possible to choose a plant trial instead of uploading a phenotype data file. Only trials with scores available to logged in user are allowed. Genotype CSV file template is generated with sample names taken from the trial data.


app/assets/javascripts/components/gwas_analysis.coffee
```diff
@@ -69,9 +84,24 @@ window.GwasAnalysis = class GwasAnalysis extends Component
...
+
+    $fileinput.removeClass('hidden')
+    $result.addClass('hidden')
+
+    this.$("#analysis_#{field.replace(/-/g, '_')}_id").val("")
+
+  updateGenotypeTemplateLink: (plant_trial_id) =>
+    $link = this.$("#analysis_gwas_genotype_template_download")
+    url = $link.attr("href")
+    url = url.replace(/&plant_trial_id=.*/, '') + "&plant_trial_id=" + plant_trial_id
```
Nuanda: This will not work if `plant_trial_id` is the first parameter and it will remove all parameters after that one. Although I see that there is no easy way to do that any better. Can you try to match `plant_trial_id` only and its value up to `&` or end of string?

app/services/analysis/gwas.rb
```diff
@@ -51,7 +71,7 @@ def store_results
     end
 
     def selected_traits
-      @selected_traits ||= analysis.args["phenos"]
+      @selected_traits ||= analysis.meta["phenos"]
```
Nuanda: Why `meta` instead of `args`?

app/controllers/analyses/data_files_controller.rb
```diff
@@ -55,4 +55,8 @@ def create_params
   def decorate_data_file(data_file)
     AnalysisDataFileDecorator.decorate(data_file)
   end
+
+  def plant_trial
+    PlantTrial.visible(current_user.id).find(params[:plant_trial_id]) if params[:plant_trial_id].present?
```
Nuanda: Maybe `find_by` and let `nil` pass to the Generator (which is able to deal with `nil`s anyway)?

app/services/analysis/gwas/plant_trial_phenotype_csv_builder.rb
```diff
@@ -0,0 +1,66 @@
...
+          csv_file.flush
+          csv_file.rewind
+        end
+      end
+
+      private
+
+      def exporter(plant_trial)
+        Submission::PlantTrialExporter.
+          new(OpenStruct.new(submitted_object: plant_trial, user: plant_trial.user))
```
Nuanda: Hmm, unexpected. But interesting. ;)

spec/features/analyses/gwas/creation_spec.rb
```diff
@@ -59,6 +66,42 @@
...
+    scenario "is successful", js: true do
+      click_link("Start!")
+      fill_in("Analysis name", with: "Some interesting name")
+
+      select(plant_trial.plant_trial_name, from: "analysis_plant_trial_id")
+
+      attach_file("genotype-data-file", fixture_file_path("gwas-genotypes.csv"))
+      attach_file("map-data-file", fixture_file_path("gwas-map.csv"))
+
+      sleep(1)
```
Nuanda: Why is this needed? For the select widget callback to reset the plant_trial_id param value?

spec/rails_helper.rb
```diff
@@ -83,14 +83,19 @@
...
         def index_document; end
         def update_document; end
         def delete_document; end
-      end.new
+        def client; FakeEsClient.new; end
+      end
 
       searchable_models.each do |model|
-        allow_any_instance_of(model).to receive(:__elasticsearch__).and_return(es)
+        allow_any_instance_of(model).to receive(:__elasticsearch__).and_return(FakeEsProxy.new)
```
Nuanda: Is this something that helps against the random search spec failures?

app/services/analysis/gwas/plant_trial_phenotype_csv_builder.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        Result.new(trait_ids, trait_id_indices, sample_ids, trait_scoring)
+      end
+
+      def build_csv(plant_trial)
+        data = build(plant_trial)
+        headers = %w(ID) + data.trait_ids.map { |trait_id| trait_id.gsub(/\s+/, ".") }
+
+        Tempfile.new(["plant-trial-phenotype", ".csv"]).tap do |csv_file|
+          csv_file.write(headers.join(",") + "\n")
```
kammerer: Note to self: use `CSV ` module as it appears that trait and sample names may include unsafe characters.

app/services/analysis/gwas.rb
kammerer: Maybe that is not the best name, but I changed it because at first I intended to keep command line args for GWASSER.R executable. Right now it holds also PT id and in the future it can possibly hold other analysis parameters.

app/controllers/analyses/data_files_controller.rb
kammerer: I think it would be slightly worse solution, because it would silently accept invalid PT id passed in params.

spec/features/analyses/gwas/creation_spec.rb
kammerer: No, it is needed for the upload to be performed, but I found more explicit way to check that.

spec/rails_helper.rb
kammerer: I simply stub another ES client method. Somehow I triggered it in a newly added spec. This is not relevant to specs which actually hit ES.

kammerer:
@Nuanda I addressed all comments one way or another.

PR #729, opened 2017-06-15
Handle invalid new accession in scoring upload for PT submission

### LINKED ISSUE #728: Plant trial submission cannot be finalized if new accession is not valid
Observed for submission #217 created by Hannah Rees on 2017-06-15.

```
An ActiveRecord::RecordInvalid occurred in submissions#update:

  Validation failed: Year produced is the wrong length (should be 4 characters)
  app/services/submission/plant_trial_finalizer.rb:47:in `find_or_create_plant_accession'
```

kammerer:
Fixes #728 

User is informed about issues with new accession after uploading the file.


Nuanda:
Are you able to write a spec for this case? Or just extend an existing one?

kammerer:
I added basic spec.

PR #731, opened 2017-06-27
First implementation of BrAPI v1 phenotypes-search call

CarlosHorro:


app/controllers/brapi/v1/phenotypes_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+        # We currently only retrieve public records. This is already done at query level
+        #if (!row[:user_id] || row[:published] )
+        
+        intermediate_hash_id = row["observationUnitDbId"]
+        existing_row = intermediate_hash[intermediate_hash_id]
+        
+        # We introduce data related to plot numbers, plants, etc
+        begin
+          observation_levels_hash = JSON.load(row.delete("observation_levels_json"))
+          if observation_levels_hash != nil
```
kammerer: Comparison with `nil` is superfluous as `nil` itself is falsy.

app/controllers/brapi/v1/phenotypes_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+          observationTimeStamp: row.delete("observations_observationTimeStamp"),
+          season: row.delete("observations_season"),
+          collector: row.delete("observations_collector"),            
+          value: row.delete("observations_value")
+        }
+        
+        if existing_row.nil? 
+          observations = []
+          observations << observation_hash
+          row["observations"] = observations
```
kammerer: This line and the two above could be replaced with `row["observations"] = [observations_hash]`

app/controllers/brapi/v1/phenotypes_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+      
+      json_response = { 
+        metadata: json_metadata(page_size, page, total_count, total_pages),
+        result: {
+          data: json_result_array
+        }
+      }
+     
+      render json: json_response, except: ["id", "user_id", "created_at", "updated_at", "total_entries_count"]
+
```
kammerer: Why add excessive vertical whitespace in such places? Language keywords are meaningful enough.

app/controllers/brapi/v1/phenotypes_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+  private
+
+  
+  def get_page
+    return (params[:page].to_i != 0)?params[:page].to_i : 1
+  end
+  
+  def get_page_size
+    return (params[:pageSize].to_i != 0)?params[:pageSize].to_i : 1000
```
kammerer: On the other hand, more whitespace between operators and operands improves readability (I already linked relevant part of Ruby style guide before).

spec/requests/brapi/v1/phenotypes_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+              expect(observationXref[key]).to be_nil.or(be_a String)
+            else
+              raise_error('Unexpected property name: observationXref['+key+']')
+            end
+          end  
+        end
+      end
+      
+      expect(result['observations']).to be_a Array             # mandatory field
+      if (result['observations'].is_a? Array)
```
kammerer: It is already established that the value is an array in the line above.

spec/requests/brapi/v1/phenotypes_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+        observations = result['observations']
+        
+        expect(observations.size).to be > 0
+        if (observations.size > 0)
```
kammerer: It is already established that the size is greater than 0 in the line above.

kammerer:
I will continue with the review tomorrow.

CarlosHorro:
About the documentation, we have a working document still in progress with our BIP-BrAPI mappings here:
https://docs.google.com/spreadsheets/d/1UicLL_FghRc8X-rVpAwlDfxQdJoRaoI4RR6icpN7J-E/edit?usp=sharing
Do I include it into the code?

spec/requests/brapi/v1/phenotypes_spec.rb
CarlosHorro: I have a doubt. If that is not an array, would next lines be executed without the 'if clause'? 
...
observations = result['observations']
        
        expect(observations.size).to be > 0
        if (observations.size > 0)
          observation = observations[0]
          
          expect(observation['observationVariableDbId']).to be_a String     # mandatory field
  
          expect(observation).to be_nil.or(be_a Hash)
          observation.keys.each do |key|
            if (['observationDbId','observationVariableName','season','observationTimeStamp',
                'collector','observationVariableDbId'].include? key)
              expect(observation[key]).to be_nil.or(be_a String)
            else
              raise_error('Unexpected property name: observation['+key+']')
            end
          end  
        end

kammerer:
Thanks for the link. I think it would be good to have that documentation included in the repository or at least linked from the README.md at some point (before BrAPI branch is merged to master).

spec/requests/brapi/v1/phenotypes_spec.rb
kammerer: If it is not an array then the expectation in line 135 will fail and nothing more will be executed.

app/controllers/brapi/v1/phenotypes_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+      params['germplasmDbIds'], params['observationVariableDbIds'], params['studyDbIds'], 
+      params['locationDbIds'], params['seasonDbIds'], 
+      params['sortBy'], params['sortOrder'], page, page_size, false)
+    
+    records = result_object.values
+   
+    if records.nil? || records.size ==0
+      render json: { reason: 'Resource not found' }, status: :not_found  # 404
+    else
+      json_result_array = []
```
kammerer: I think this part could use some restructuring in order to improve readability. Perhaps something like below? This is just a proposition though.

```ruby
observation_units = []
observations = Hash.new { |h, k| h[k] = [] }

result_object.each do |observation|
  observation_unit_id = observation['observationUnitDbId']
  observation_unit = extract_observation_unit(observation)

  if observations[observation_unit_id].empty?
    observation_unit['observations'] = observations[observation_unit_id]
    observation_units << observation_unit
  end 

  # handle design factor transformations...

  observations[observation_unit_id] << observation
end
```

kammerer:
Except for the JSON parsing error which IMO cannot be handled like this, please treat my comments as suggestions not requests of changes.

app/controllers/brapi/v1/phenotypes_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+        # We introduce data related to plot numbers, plants, etc
+        begin
+          observation_levels_hash = JSON.load(row.delete("observation_levels_json"))
+          if observation_levels_hash != nil
+              row["plotNumber"]= observation_levels_hash["plot"]
+              row["plantNumber"]= observation_levels_hash["plant"]
+              row["blockNumber"]= observation_levels_hash["block"]
+              row["observationLevel"]= observation_levels_hash.keys[observation_levels_hash.keys.length-1]
+          end
+        rescue JSON::ParserError => e
```
CarlosHorro: observation_levels_json is based on design_factors.design_factors, which is free text introduced by the user, so it's not totally reliable.
In this case the exception only indicates the data was introduced in a different format than I need to manage it, but I think that format is not (still) mandatory when user introduces it, so I think I must not stop the normal execution of the procedure .

kammerer: If ORM was used this wouldn't even be an issue.

app/controllers/brapi/v1/phenotypes_controller.rb
```diff
@@ -17,100 +17,95 @@ def search
...
+    observation_hash = {
+      observationDbId: observation_row.delete("observations_observationDbId"),
+      observationVariableDbId: observation_row.delete("observations_observationVariableDbId"),
+      observationVariableName: observation_row.delete("observations_observationVariableName"),
+      observationTimeStamp: observation_row.delete("observations_observationTimeStamp"),
+      season: observation_row.delete("observations_season"),
+      collector: observation_row.delete("observations_collector"),            
+      value: observation_row.delete("observations_value")
+    }
+    return observation_hash
```
kammerer: No need to assign local variable and return explicitly.

app/controllers/brapi/v1/phenotypes_controller.rb
```diff
@@ -17,100 +17,95 @@ def search
...
+    # We introduce data related to plot numbers, plants, etc
+    begin
+      observation_levels_hash = JSON.load(observation_row.delete("observation_levels_json"))
+      if !observation_levels_hash.nil?
+          observation_row["plotNumber"]= observation_levels_hash["plot"]
+          observation_row["plantNumber"]= observation_levels_hash["plant"]
+          observation_row["blockNumber"]= observation_levels_hash["block"]
+          observation_row["observationLevel"]= observation_levels_hash.keys[observation_levels_hash.keys.length-1]
+      end
+    rescue JSON::ParserError => e
```
kammerer: You are still ignoring this error. This is not OK.

app/services/brapi/v1/germplasm_queries.rb
```diff
@@ -122,17 +139,23 @@ def execute_statement(sql, atts)
...
+      begin
+        if( (atts != nil) && !(atts.empty?) )
+          results = @connection.exec_prepared("brapi_statement", atts)
+        else
+          results = @connection.exec_prepared("brapi_statement")
+        end
+      rescue PG::Error => e
+        @connection.exec("ROLLBACK") 
+        results = nil
+        Rails.logger.warn { "Encountered an error executing a BrAPI germplasm-related query: #{e.message} #{e.backtrace.join("\n")}" }
```
kammerer: We have error notification in place. It should be used here in addition to logging.

CarlosHorro: I'm retrieving the error notification to the user in the controller when I receive the results = nil : 
if records.nil? || records.size ==0
        render json: { reason: 'Resource not found' }, status: :not_found  # 404
      else

Is it ok? if not, could you give me an example, please?  I was looking for examples in services but I only found things like
...
rescue Exception => e
      message = "Problem accessing public ORCiD API for uid #{uid}."
      message += " Reason: \"#{e.message}\"."
      message += ' No public user information available for the moment.'
      Rails.logger.warn "WARNING: #{message}"
      return { status: :error, message: message }
    end
...
in orcid_client.rb (which I'm not sure if it's generalizable)

kammerer: There are several issues with this approach. Firstly, the query failing in such way is not something we can ignore, this is clearly error in application logic and we want to know about it (well, it could be a timeout but we still want to know about it). We use mail notifier which sends information about exceptions that are not handled at the controller level ([smartinez/exception_notification](https://github.com/smartinez87/exception_notification) gem). However it will not run if the exception is handled by application code. In some such cases it makes sense to call it directly:
```
ExceptionNotifier.notify_exception(e)
```

Secondly, returning 404 response in case of application error is misleading. 500 should be returned so that the user knows this is in fact an error and not that we have no information he/she requested. 

Lastly, returning `nil` from method to indicate error may not be terribly wrong but it could be done in more explicit way, e.g.: do cleanup, log debugging info, send error notification and return `:error` instead of `nil` or alternatively do cleanup and re-raise the exception and use error handler at the controller level (`rescue_from PG::Error`). Optionally, wrap `PG::Error` with custom `QueryError` to avoid leaky abstraction. Where possible error handling should be separated from other logic as it tends to make code less readable.

```ruby
class Brapi::V1::GermplasmQueries
# ...
  rescue PG::Error => ex
    @connection.exec("ROLLBACK") 
    raise QueryError, sql # or whatever useful information (note that previous exception is automatically attached as the cause of new exception)
  ensure
    @connection.exec("DEALLOCATE brapi_statement") 
  end
# ...
```

```ruby
class Brapi::V1::ApidocsController < ApplicationController
# ...
  rescue_from Brapi::QueryError do |ex|
    logger.warn { ... }
    ExceptionNotifier.notify_exception(ex, env: request.env, data: { sql: ex.sql, cause: ex.cause.message })
    return json: { reason: 'Internal error', message: 'There was some error managing studies/search query' }, status: :internal_server_error
  end
# ...
```

kammerer: Edit: I see that you actually return 500, this is tested before the condition you mentioned. Oh, the code I attached is not tested, it is only meant as an example.

CarlosHorro: Thank you very much for your guidance. I've being doing some tests about this with studies/id call and I've created a new custom exception QueryError into new module 'Errors::Brapi' (app/errors is now included into autoload_paths).
If the new approach is correct I would extend it into all other calls.

kammerer: Personally, I would rather put the exception class into `app/services/brapi` even though it technically is not a service. We did not use custom exceptions before and I don't think it makes sense to add new directory hierarchy just for this on class.

Another minor thing - it would probably be OK to put `rescue_from` invocations in `Brapi::BaseController` instead of specific controllers.

PR #737, opened 2017-08-16
Brapi v1 swagger

CarlosHorro:
All nodes are trying to expose Swagger interfaces to test APIs, and this would be our implementation of Swagger 2.0 using swagger-blocks gem.
Entry point would be 
bip.earlham.ac.uk/swagger-apidocs/brapi/v1/index.html
and json information is automatically generated from code at
bip.earlham.ac.uk/brapi/v1/apidocs

This first implementation covers first Brapi v1, 5 calls:
/germplasm-search
/phenotypes-search
/studies-search
/studies/:id
/studies/:studyDbId/germplasm  (pending on being merged in another branch)


CarlosHorro:
I didn't know that gem... I'm testing it and seems ok, although there are some problems I cannot totally solve, as issues managing multiple versions. This works in the example they provide, but when I try the same, it always returns this error when I provide more than 1 version to the config file:

Can't read swagger JSON from http://localhost:3000/swagger-apidocs2/{:v1=&gt;&quot;/../brapi/v1/apidocs&quot;, :v2=&gt;&quot;/../brapi/v1/apidocs&quot;}

instead providing the intermediate page with both versions.
However, it seems that we will only provide one swagger API so maybe that is not very important.

Another minor issue its related with swagger-blocks, which it doesn't allow me to generate its json in a different location than the one configured in routes.rg (brapi/v1/apidocs). (:host and :basePath keys are ignored). I would like to keep current controller in its location as it separates clearly Brapi V1 swagger from any other future swagger implementation.
Furthermore, swagger_ui_engine always looks for jsons into its main folder, so, in order to use swagger-blocks jsons by swagger-ui-engine, I can only write something like:

config.swagger_url = '/../brapi/v1/apidocs'

Which it's not very nice but works...

So, if these are not considered as important problems, I can replace old swagger ui things by new swagger_ui_engine configurations...

kammerer:
It appears that specifying multiple sources is a feature from the latest version of the engine. This version however is dependent on Rails 5 and we are still using Rails 4, so older version was installed by bundler.
The latter thing does not seem to be a problem to me, actually it has the benefit that you always see docs from your app instance.

All in all, if you are OK with using the engine, it works and it does not limit you, I would say it is better to do that instead of bundling the UI into `/public`. Otherwise, you may stick to current solution.

CarlosHorro:
I agree with you and I think it's a good improvement. I've updated the branch with the new changes.

CarlosHorro:
Before merging this pull request, I would need the OK to the pull request related to studies/:studyDbId/germplasm (https://github.com/TGAC/brassica/pull/738) as this Swagger implementation includes documentation about that call, please.

PR #738, opened 2017-08-16
First implementation of Brapi call: studies/:studyDbId/germplasm

CarlosHorro:


CarlosHorro:
Um... as now there are some things in the Brapi v1 branch (QueryError mainly) which will have to be used by this one, I think I will cancel this pull request and merge this local branch with the local branch brapi v1 in order to include everything into it.

PR #741, opened 2017-08-24
Support for HapMap files in GWAS

### LINKED ISSUE #727: Hapmap files as GWAS input

kammerer:
Fixes #727 

Example file can be found in fixtures: `spec/fixtures/files/gwas-genotypes.hapmap` (it matches `spec/fixtures/files/gwas-phenotypes.csv`).


app/services/analysis/gwas/genotype_hapmap_parser.rb
```diff
@@ -0,0 +1,87 @@
...
+        def valid?
+          errors.empty?
+        end
+
+        def sample_ids
+          @sample_ids ||= columns[HEADER_COLUMNS.size..-1]
+        end
+
+        def each_record(&blk)
+          enum = io.each_line.lazy.map { |line| Record.new(sample_ids, line.split(/\s+/)) }
```
Nuanda: Can we have a more descriptive name instead of `enum`?

app/services/analysis/gwas/genotype_hapmap_to_csv_converter.rb
```diff
@@ -0,0 +1,147 @@
...
+      #   0 for no mutation
+      #   NA if call could not be made
+      def sample_value(mutation, sample)
+        verify_sample(sample)
+
+        _original, alternative = mutation.split("/")
+
+        return "NA" if alternative == "NN"
+
+        alternative_count = sample.split("").count { |allele| allele == alternative }
```
Nuanda: I must say I am a bit surprised you need to use `split` in such a strange way to turn a string into an enumerable. But it looks like there really is no other (simple) way... :)

app/views/analyses/new/_genotype_data_file_help.html.haml
```diff
@@ -1,10 +1,11 @@
...
-  Genotype data needs to be provided in one of two accepted formats -
-  #{link_to 'VCF', 'http://samtools.github.io/hts-specs/', title: 'Variant Call Format', target: "_blank"} or
+  Genotype data needs to be provided in one of the accepted formats -
+  #{link_to 'VCF', 'http://samtools.github.io/hts-specs/', title: 'Variant Call Format', target: "_blank"},
+  #{link_to 'HapMap', 'http://augustogarcia.me/statgen-esalq/Hapmap-and-VCF-formats-and-its-integration-with-onemap/#hapmap', title: "HapMap", target: "_blank"} or
   #{link_to 'CSV', 'https://en.wikipedia.org/wiki/Comma-separated_values', title: 'Comma-separated values', target: "_blank"}.
   %br
   %br
-  VCF is a well-defined format and any such file should be acceptable. However, for best performance low quality
+  VCF and HapMap are well-defined formats and any such files should be acceptable. However, for best performance low quality
```
Nuanda: Can you say "any such files"? "Any such file" maybe?

Nuanda:
I have a spec problem on my side - when running specs in `spec/features/analyses/gwas/` I get the following error:

```
     ActionView::Template::Error:
       TypeError: Cannot set property 'root' of null
         (in ...app/assets/javascripts/data_grid_specific_columns.js.coffee)
```
two times. The other is for `manhattan_plot.coffee`. Are you able to help me with fixing that?

app/services/analysis/gwas/genotype_hapmap_to_csv_converter.rb
kammerer: It appears there is a better way: `sample.chars`.

app/services/analysis/gwas/genotype_hapmap_parser.rb
kammerer: Changed to `records`.

app/views/analyses/new/_genotype_data_file_help.html.haml
kammerer: Fixed.

kammerer:
@Nuanda I did not encounter the error you mention but I googled it and found that updating coffeescript should help, so I did that. Can you test it again?

Nuanda:
It does work ok now.

PR #745, opened 2017-09-25
Taxonomy terms ncbi migration

CarlosHorro:
Migration file and csv file with NCBI - GR mapping.


Nuanda:
@CarlosHorro Carlos, @kammerer will review this after he's back from his leave -- beginning of October.

kammerer:
@CarlosHorro In the future you can make small PRs as such directly against master branch.

db/schema.rb
```diff
@@ -57,6 +57,7 @@
   end
 
   add_index "api_keys", ["token"], name: "api_keys_token_idx", unique: true, using: :btree
+  add_index "api_keys", ["token"], name: "index_api_keys_on_token", using: :btree
```
kammerer: Please, do not commit any changes to `db/schema.rb` which are not related to the migration you added.

db/taxonomy_terms_ncbi_updated.csv
```diff
@@ -0,0 +1,189 @@
+ID,LABEL,New_Label,ncbi_taxon,NAME,TAXONOMY_TERM_ID,CREATED_AT,UPDATE_AT,CANONICAL
```
kammerer: `db/data/taxonomy_terms_ncbi_updated.csv` would be a slightly better location.

db/schema.rb
CarlosHorro: Sorry, I don't know why these changes were generated automatically.

PR #748, opened 2017-10-04
Support for Biosamples Ids and migration of currently uploaded samples with their ids

CarlosHorro:
Following last suggestion by kammerer I've made this little PR directly to master.


app/models/plant_scoring_unit.rb
```diff
@@ -7,6 +7,7 @@ class PlantScoringUnit < ActiveRecord::Base
   has_many :trait_scores, dependent: :destroy
 
   validates :scoring_unit_name, :plant_trial_id, :plant_accession_id, presence: true
+  validates :biosamples_id, uniqueness: true
```
kammerer: You need to add `allow_nil: true` to make it work. Check specs.

PR #749, opened 2017-10-05
updated About Us and About BIP pages

wjurkowski:
I hope formatting of paragraphs is as it should be


kammerer:
I fixed a few syntax issues.

PR #757, opened 2017-12-05
text updates in population submission wizard

wjurkowski:
Update adds information about PP template in step zero. This was requested by the users to see the  template upfront


kammerer:
I fixed formatting and deployed it to production.

PR #766, opened 2017-12-28
Add trait admin UI

### LINKED ISSUE #765: Trait admin UI
Introduce a possibility for a BIP administrator superuser to submit new
Traits to BIP (in form of a simple UI) and have them available to the users.

kammerer:
Fixes #765 

I skipped trait removal on purpose.


app/views/admin/traits/_form.html.haml
```diff
@@ -0,0 +1,33 @@
+= form_for([:admin, @trait]) do |f|
+  = render partial: "/errors", locals: { errors: @trait.errors }
+
+  .form-group
+    = f.text_field :name, required: true
```
Nuanda: I made an interesting observation. When created trait A, and created a PT submission with a TD using this trait A, and then changing A's name to B, the saved submission retains 'A' as the name of the trait. This has some additional consequences, like, for instance, now I can create another trait named A and have another TD, in the same submission, with the same trait name, though it is not the same trait (technically).

I'm not sure if this needs to be addressed or not, but I decided to describe it here just in case.

kammerer: I have not noticed anything like you describe. It is possible thought that caching might make it look so in some places in the UI. Where exactly did you see this effect?

app/views/admin/traits/_form.html.haml
```diff
@@ -0,0 +1,33 @@
...
+
+  .form-group
+    = f.text_field :label, required: true, help: "Eg. ontological term or 'BIP/EI' for traits not defined in any ontology"
+
+  .form-group
+    = f.text_area :description
+
+  .form-group
+    = f.label :canonical
```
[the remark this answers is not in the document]
kammerer: I think it won't hurt to leave it but I added help text to make its meaning clear.

app/views/admin/traits/_form.html.haml
Nuanda: After I created a TD during a PT submission, and saved&exited from the submission, this submission listed the A name in its scored trait list. I went to the new traits edit view and changed A into something else. Then I revisited the saved submission and it still showed A in the scored trait list (I mean the form-list where newly defined TDs are added). I didn't test anything past that point, e.g. if the generated scoring template has the old trait name.

I suspect some temporary lists stored inside the submission json might have copied the A name and retained it even after that was changed.

kammerer: Thanks, you are right about the temporary list. I think we can live with the fact that the name in this list does not reflect changes, after all the person editing trait name should do so responsibly and not change it to something unrelated. However, there is a technical issue I need to fix - new TD references Trait by name and it results in an error during finalization of the submission if Trait is not found.

kammerer: Another thing I did not think about earlier - changing Trait's name will mess up parsing of scoring CSV. This may be tricky to handle, and since the request was mainly for adding new Traits I am considering complete removal of Trait editing functionality or freezing the name.

Nuanda: Can you consider checking if a given trait is used in any TD? This should be simple and then you can block any name edit. I think you will need it anyway if you consider adding trait removal in the future.

kammerer: This won't help much, as TD is actually created only when submission is finalized. So, a Trait can be already chosen for a new TD in submission while the TD actually does not exist yet.

Implementing Trait removal/name editing probably requires checking submission content. For now, I am blocking name editing. I may do something more about this if time permits.

PR #778, opened 2018-01-31
Datatable navigation extensions

### LINKED ISSUE #648: Add PL <-> PA, PV <-> PA and PL -> TS navigation possibility to data tables

kammerer:
Fixes #648 

This is about linking PL -> PA and PV -> PA (everything else is already implemented). The former is pretty straightforward as this is a normal `has_many` relation. The latter however is somewhat tricky and needs to be handled in a non-standard way as the relation `PlantVariety#plant_accessions` does not include all PAs we want to show/count for a PV.

I created a read-only model `PlantVarietyAccession` based on a database view which allows for easier access to PAs both directly and indirectly related to PV. I then used this model/db view to implement `PlantVariety#table_data` and show count of all related PAs. Unfortunately, there is no easy way to use counter cache column with this approach, so the counter attribute is not cached.
 
I am not happy about naming the new relation `PlantVariety#plant_variety_accessions`, I feel that it would make more sense to swap the name with already existing `PlantVariety#plant_accessions`. However, if I did this then the name of the new model/db view would be somewhat confusing...

### DIFF
--- a/app/models/concerns/publishable.rb
+++ b/app/models/concerns/publishable.rb
@@ -3,38 +3,13 @@ module Publishable
 
   included do
     include ActiveModel::Validations
+    include PublishableQueries
 
     validates_with PublicationValidator
-
-    scope :published, -> { where(published: true) }
-    scope :not_published, -> { where(published: false) }
-    scope :visible, ->(uid = nil) {
-      condition = "#{table_name}.user_id IS NULL OR #{table_name}.published = TRUE"
-      condition += " OR #{table_name}.user_id = #{uid}" if uid
-      where condition
-    }
-
-    scope :hidden, ->(uid = nil) {
-      condition = "#{table_name}.published = FALSE"
-      condition += " AND #{table_name}.user_id <> #{uid}" if uid
-      where condition
-    }
 
     before_validation -> {
       self.published_on ||= Time.now if published?
     }
-
-    def revocable?
-      published? && Time.now < revocable_until
-    end
-
-    def revocable_until
-      published_on + 1.week if published?
-    end
-
-    def private?
-      !(published? || user.nil?)
-    end
 
     def publish
       return if published?
--- a/app/models/concerns/publishable_queries.rb
+++ b/app/models/concerns/publishable_queries.rb
@@ -1,0 +1,31 @@
+module PublishableQueries
+  extend ActiveSupport::Concern
+
+  included do
+    scope :published, -> { where(published: true) }
+    scope :not_published, -> { where(published: false) }
+    scope :visible, ->(uid = nil) {
+      condition = "#{table_name}.user_id IS NULL OR #{table_name}.published = TRUE"
+      condition += " OR #{table_name}.user_id = #{uid}" if uid
+      where condition
+    }
+
+    scope :hidden, ->(uid = nil) {
+      condition = "#{table_name}.published = FALSE"
+      condition += " AND #{table_name}.user_id <> #{uid}" if uid
+      where condition
+    }
+
+    def revocable?
+      published? && Time.now < revocable_until
+    end
+
+    def revocable_until
+      published_on + 1.week if published?
+    end
+
+    def private?
+      !(published? || user.nil?)
+    end
+  end
+end
--- a/app/models/concerns/relatable.rb
+++ b/app/models/concerns/relatable.rb
@@ -18,8 +18,8 @@ module Relatable extend ActiveSupport::Concern
       end
     end
 
-    def self.privacy_adjusted_count_columns
-      count_columns.map do |column|
+    def self.privacy_adjusted_count_columns(except: [])
+      (count_columns - except).map do |column|
         counter_column = column.split(/ as /i)[0]
         as_column = column.split(/ as /i)[-1]
         "(#{table_name}.#{counter_column} - coalesce(#{get_related_model(counter_column)}.hidden, 0)) AS #{as_column}"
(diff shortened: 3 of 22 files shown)


app/models/concerns/relatable.rb
```diff
@@ -18,8 +18,8 @@ def self.counter_names
       end
     end
 
-    def self.privacy_adjusted_count_columns
-      count_columns.map do |column|
+    def self.privacy_adjusted_count_columns(except: [])
+      (count_columns - except).map do |column|
```
Nuanda: This concern is rather widely used throughout the model. Can you provide some specs to test the new behavior?

app/models/plant_line.rb
```diff
@@ -74,6 +75,12 @@ def self.permitted_params
     ]
   end
 
+  def self.count_columns
+    [
+      'plant_accessions_count'
```
Nuanda: Either this, or CoffeeScript's ``baseColumnDefs`` and ``createCounterLink``, as when I have a PL with no accessions, I get a "NULL plant accession" clickable link under the "Related" button:

spectacle b18623

(PLName: Varuna). I should, however, rather get a proper unclickable link, like this:

spectacle h18623

When there are >0 PAs associated, the links looks to be correct. I, however, found other oddities in the PA table, like:

spectacle e18623

PAName: rres_GK090856

The counting migration didn't report any problem for me. Could it be it's some kind of older problem? In production this PA has a correct count of 12 PSUs.

Nuanda: After removing my cache it looks like the first problem is gone, but this second one stays. But maybe it's just me. Simply try to check again such specific cases and see if that happens to you as well.

app/models/concerns/publishable.rb
```diff
@@ -3,6 +3,7 @@ module Publishable
 
   included do
     include ActiveModel::Validations
+    include PublishableQueries
```
Nuanda: This redefines some scopes (published, not_published, visible, hidden) which are also defined inside ``Publishable``. Shouldn't the definitions be removed from ``Publishable``?

app/models/concerns/publishable_queries.rb
```diff
@@ -0,0 +1,31 @@
...
+    scope :not_published, -> { where(published: false) }
+    scope :visible, ->(uid = nil) {
+      condition = "#{table_name}.user_id IS NULL OR #{table_name}.published = TRUE"
+      condition += " OR #{table_name}.user_id = #{uid}" if uid
+      where condition
+    }
+
+    scope :hidden, ->(uid = nil) {
+      condition = "#{table_name}.published = FALSE"
+      condition += " AND #{table_name}.user_id <> #{uid}" if uid
```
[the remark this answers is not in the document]
kammerer: Yes, it's old code. I am not sure about actual contents of the database but in model layer this heuristics is enforced by a custom validator which rejects "ownerless" records having published flag set to false.

app/models/plant_accession.rb
```diff
@@ -45,18 +46,11 @@ def self.table_data(params = nil, uid = nil)
...
-      errors.add(:plant_variety_id, :not_with_plant_line)
-    end
-  end
-
   def self.table_columns
     [
       'plant_accession',
       'plant_lines.plant_line_name',
-      'CASE WHEN plant_varieties.id IS NULL THEN plant_line_varieties.plant_variety_name ELSE plant_varieties.plant_variety_name END AS plant_variety_name',
+      'plant_varieties.plant_variety_name',
```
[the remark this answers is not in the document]
kammerer: Kind of, yes. `PlantAccession#table_data` is now based on the database view which lists accessions with corresponding `plant_variety_id` (even if accession is related to variety through a plant line).

app/models/plant_line.rb
kammerer: I think the PSU count problem is not related, not happened to me. But I will yet make some checks with data dumped from production.

app/models/concerns/publishable.rb
kammerer: Correct, I forgot to remove them.

app/models/plant_accession.rb
```diff
@@ -76,25 +70,27 @@ def self.count_columns
...
+    ]
+  end
+
   def self.permitted_params
     [
       :fetch,
       query: params_for_filter(table_columns) +
         [
+          'plant_lines.id',
+          'plant_varieties.id',
```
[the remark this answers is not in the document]
kammerer: Yup, they are fine.

Nuanda:
Thanks for answers. I think there are still two outstanding issues - the specs and the TODO.

kammerer:
I will get back to it next week.

app/models/concerns/relatable.rb
kammerer: I added some minimal tests. This is also tested indirectly by existing tests for `.table_data`.

app/models/plant_accession.rb
```diff
@@ -1,8 +1,10 @@
 class PlantAccession < ActiveRecord::Base
-  belongs_to :plant_line
+  belongs_to :plant_line, counter_cache: true
   belongs_to :plant_variety
   belongs_to :user
 
+  # TODO: add callbacks to update PlantVariety#plant_variety_accessions_count
```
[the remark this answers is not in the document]
kammerer: I removed this TODO. Since current implementation of `PlantVariety.table_data` does not seem to be slow I don't think it is that important to have this count cached at this point.

PR #779, opened 2018-02-04
Do not reference form step in submission content

kammerer:
I need to change the structure of submission forms again. Before doing that I decided to change the way submission content is stored - it won't be nested under step name anymore:

```ruby
# before
submission.content.step01.plant_trial_name

# after
submission.content.plant_trial_name
```

I don't think having this nesting benefited us anyway and removing it will save me the trouble with migration of data between the respective steps.

### DIFF
--- a/app/services/submission/plant_trial_finalizer.rb
+++ b/app/services/submission/plant_trial_finalizer.rb
@@ -25,7 +25,7 @@ class Submission::PlantTrialFinalizer
   attr_accessor :submission
 
   def create_new_trait_descriptors
-    @new_trait_descriptors = (submission.content.step02.new_trait_descriptors || []).map do |attrs|
+    @new_trait_descriptors = (submission.content.new_trait_descriptors || []).map do |attrs|
       attrs = attrs.with_indifferent_access
       attrs = attrs.merge(common_data)
       if attrs[:trait].present?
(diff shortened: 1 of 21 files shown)


app/services/submission/plant_trial_finalizer.rb
```diff
@@ -164,22 +164,26 @@ def create_scoring
...
 
   def create_plant_trial
-    attrs = submission.content.step01.to_h.merge(common_data)
-    design_factor_names = submission.content.step04.design_factor_names || []
+    attr_names = [:plant_trial_name, :project_descriptor, :plant_population_id, :plant_trial_description, :trial_year,
+                  :institute_id, :country_id, :trial_location_site_name, :place_name, :latitude, :longitude, :altitude,
+                  :terrain, :soil_type, :statistical_factors, :data_status]
 
-    if plant_population = PlantPopulation.find_by(id: submission.content.step01.plant_population_id)
+    attrs = submission.content.to_h.slice(*attr_names).merge(common_data)
```
Nuanda: In `create_plant_population` you merge submission content to common_data. Here, you proceed in a reverse fashion. Any particular reason?

kammerer: Not really, it's because of the way code was structured before these changes.

kammerer:
Thanks!

